PR #1, opened 2018-06-04
Implement ToScalar from spec, and small refactor to match spec more closely

str4d:
Test vectors from sapling_key_components.py and sapling_signatures.py are not
altered by this commit.


daira:
utACK

PR #2, opened 2018-06-04
Refactor error handling

daira:


str4d:
First commit is already in master, but git should work it out when merging this.

PR #4, opened 2018-06-05
Implement Zcash test vector output format

str4d:
Usage: `python3 sapling_foo.py -t zcash`

### DIFF
--- a/sapling_generators.py
+++ b/sapling_generators.py
@@ -3,7 +3,7 @@ from binascii import hexlify
 from pyblake2 import blake2s
 
 from sapling_jubjub import Point, JUBJUB_COFACTOR
-from sapling_utils import chunk
+from tv_output import chunk
 
 # First 64 bytes of the BLAKE2s input during group hash.
 # This is chosen to be some random string that we couldn't have
--- a/sapling_generators.py
+++ b/sapling_generators.py
@@ -1,9 +1,8 @@
 #!/usr/bin/env python3
-from binascii import hexlify
 from pyblake2 import blake2s
 
 from sapling_jubjub import Point, JUBJUB_COFACTOR
-from tv_output import chunk
+from tv_output import tv_rust
 
 # First 64 bytes of the BLAKE2s input during group hash.
 # This is chosen to be some random string that we couldn't have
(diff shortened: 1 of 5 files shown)


bitcartel:
Here is the output I get.  How are the test vectors to be used?  Is there some specific C++ code we use?  With key value pairs we could use Univalue.
```
python3 sapling_key_components.py -t zcash
[
    ["From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/sapling_key_components.py"],
    ["sk, ask, nsk, ovk, ak, nk, ivk, default_d, default_pk_d, note_v, note_r, note_cm, note_pos, note_nf"],
    ["0000000000000000000000000000000000000000000000000000000000000000", "06880e0df04583674f05d25dcf1119cf18f84420407823aa47a53e474aa14885", "056e5e74ac7e2d26018533275e42b081f53133ecb6eaeaf01cb60bdda04e1130", "3bd696b9e13851865ee47c1d03acb5034e224d6e4fa4ab7c17049bd91369d198", "2016f18efa0efd770776328095bad71f793a5d8c58c298303e27e10f38ec44f3", "bad63fbd78a919fb88fb25a7270e04302d067bac19153c388386e5f2779ecff7", "04924751145f0faa2d9d71a5549d563eb145e22e50a9add7dfcb03edd07c0bb7", "f19d9b797e39f337445839", "1574ae41e876d7e3149fc2d3265155a945c46765f131a18cebf7c4aab0d24cdb", 0, "0000000000000000000000061536ec550286898e778dcc0e98e4ac39ac6d1739", "396407ddf23e088f270608ce4fd4fec95018c0bcc2c614b97ed5703215f93ccb", 0, "6340ab472775d96762e77d00462360bf5e1d868fa2439ca19fecfd4f56d6fa44"],
```

str4d:
@bitcartel this is the same format that all our other (mostly Bitcoin-inherited) test vectors are in. See `src/test/data` and the tests that use them.

bitcartel:
Thanks, I figured as much.  Maybe in future we could add an option to output JSON, so that the test vector data file can be easily shared, be language independent and third parties can use whatever JSON parser they have.

arcalinea:
These work now

str4d:
@bitcartel yeah, that would be simple to do (just re-add the `json` type and call `tv_json(filename, parts, vectors, False)`). But let's leave that for later.

PR #5, opened 2018-06-15
ZIP 143 test vectors

str4d:
The Zcash output is almost identical to the existing `sighash.json` output, except that amounts are provided.


str4d:
Here is the patch required to run these test vectors in place of `sighash.json`:

```diff
diff --git a/src/test/sighash_tests.cpp b/src/test/sighash_tests.cpp
index a73bd6f..ee71af6 100644
--- a/src/test/sighash_tests.cpp
+++ b/src/test/sighash_tests.cpp
@@ -281,6 +281,7 @@ BOOST_AUTO_TEST_CASE(sighash_from_data)
 
         std::string raw_tx, raw_script, sigHashHex;
         int nIn, nHashType;
+        CAmount amount;
         uint32_t consensusBranchId;
         uint256 sh;
         CTransaction tx;
@@ -292,13 +293,15 @@ BOOST_AUTO_TEST_CASE(sighash_from_data)
           raw_script = test[1].get_str();
           nIn = test[2].get_int();
           nHashType = test[3].get_int();
-          consensusBranchId = test[4].get_int();
-          sigHashHex = test[5].get_str();
+          assert(-1 == NOT_AN_INPUT);
+          amount = test[4].get_int64();
+          consensusBranchId = test[5].get_int();
+          sigHashHex = test[6].get_str();
 
           uint256 sh;
           CDataStream stream(ParseHex(raw_tx), SER_NETWORK, PROTOCOL_VERSION);
           stream >> tx;
-
+/*
           CValidationState state;
           if (tx.fOverwintered) {
               // Note that OVERWINTER_MIN_CURRENT_VERSION and OVERWINTER_MAX_CURRENT_VERSION
@@ -321,7 +324,7 @@ BOOST_AUTO_TEST_CASE(sighash_from_data)
               BOOST_CHECK_MESSAGE(CheckTransactionWithoutProofVerification(tx, state), strTest);
               BOOST_CHECK(state.IsValid());
           }
-
+*/
           std::vector<unsigned char> raw = ParseHex(raw_script);
           scriptCode.insert(scriptCode.end(), raw.begin(), raw.end());
         } catch (...) {
@@ -329,7 +332,7 @@ BOOST_AUTO_TEST_CASE(sighash_from_data)
           continue;
         }
 
-        sh = SignatureHash(scriptCode, tx, nIn, nHashType, 0, consensusBranchId);
+        sh = SignatureHash(scriptCode, tx, nIn, nHashType, amount, consensusBranchId);
         BOOST_CHECK_MESSAGE(sh.GetHex() == sigHashHex, strTest);
     }
 }
```

transaction.py
```diff
@@ -0,0 +1,185 @@
...
+
+        ret += struct.pack('<I', self.nLockTime)
+        if isOverwinterV3:
+            ret += struct.pack('<I', self.nExpiryHeight)
+
+        if self.nVersion >= 2:
+            ret += struct.pack('b', len(self.vJoinSplit))
+            for jsdesc in self.vJoinSplit:
+                ret += bytes(jsdesc)
+            if len(self.vJoinSplit):
```
daira: I suggest ``len(self.vJoinSplit) > 0`` for clarity (I know it's equivalent).

zip_0143.py
```diff
@@ -0,0 +1,164 @@
...
+def getHashJoinSplits(tx):
+    digest = blake2b(digest_size=32, person=b'ZcashJSplitsHash')
+    for jsdesc in tx.vJoinSplit:
+        digest.update(bytes(jsdesc))
+    digest.update(tx.joinSplitPubKey)
+    return digest.digest()
+
+
+# Currently assumes the nHashType is SIGHASHALL
+# and that there are no joinSplits
```
daira: Is this comment correct? It does seem to implement hashing for JoinSplits.

zip_0143.py
```diff
@@ -0,0 +1,164 @@
...
+def signature_hash(scriptCode, tx, nIn, nHashType, amount, consensusBranchId):
+    hashPrevouts = b'\x00'*32
+    hashSequence = b'\x00'*32
+    hashOutputs = b'\x00'*32
+    hashJoinSplits = b'\x00'*32
+
+    if not nHashType & SIGHASH_ANYONECANPAY:
+        hashPrevouts = getHashPrevouts(tx)
+
+    if (not nHashType & SIGHASH_ANYONECANPAY) and \
```
daira: This is correct, but write ``not (nHashType & SIGHASH_ANYONECANPAY)`` so that readers don't have to look up the precedence :-)

transaction.py
```diff
@@ -0,0 +1,185 @@
...
+            struct.pack('<Q', self.vpub_old) + \
+            struct.pack('<Q', self.vpub_new) + \
+            self.anchor + \
+            b''.join(self.nullifiers) + \
+            b''.join(self.commitments) + \
+            self.ephemeralKey + \
+            self.randomSeed + \
+            b''.join(self.macs) + \
+            bytes(self.proof) + \
+            b''.join(self.ciphertexts)
```
daira: Style nit: prefer parens to line continuations, i.e.
```
return (struct.pack(...) +
        struct.pack(...) +
        ... +
        b''.join(self.ciphertexts))
```
or
```
return (
    struct.pack(...) +
    struct.pack(...) +
    ... +
    b''.join(self.ciphertexts)
)
```

zip_0143.py
```diff
@@ -0,0 +1,164 @@
...
+        tx = Transaction(rand, OVERWINTER_TX_VERSION)
+        scriptCode = Script(rand)
+        nIn = rand.u8() % (len(tx.vin) + 1)
+        if nIn == len(tx.vin):
+            nIn = NOT_AN_INPUT
+        nHashType = SIGHASH_ALL if nIn == NOT_AN_INPUT else rand.a([
+            SIGHASH_ALL,
+            SIGHASH_NONE,
+            SIGHASH_SINGLE,
+        ])
```
daira: This doesn't test ``SIGHASH_ANYONECANPAY``.

zip_0143.py
```diff
@@ -0,0 +1,164 @@
...
+        amount = rand.u64() % MAX_MONEY
```
daira: This is very pedantic: ``rand.u64() % (MAX_MONEY+1)``.

(This would mainly act as documentation that the range is 0..MAX_MONEY inclusive.)

transaction.py
```diff
@@ -0,0 +1,185 @@
...
+            struct.pack('<I', self.nSequence)
+
+
+class TxOut(object):
+    def __init__(self, rand):
+        self.nValue = rand.u64() % MAX_MONEY
+        self.scriptPubKey = Script(rand)
+
+    def __bytes__(self):
+        return struct.pack('<Q', self.nValue) + bytes(self.scriptPubKey)
```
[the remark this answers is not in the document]
str4d: Yep, that's already being added by `Script.__bytes__()`.

zip_0143.py
str4d: Oops, forgot to update it

zc_utils.py
```diff
@@ -0,0 +1,19 @@
+import struct
+
+def write_compact_size(n):
+    if n < 253:
+        return struct.pack('B', n)
+    elif n < 0xFFFF:
```
daira: This is not correct; it is `<= 0xFFFF` (see https://bitcoin.org/en/developer-reference#compactsize-unsigned-integers ).

zc_utils.py
```diff
@@ -0,0 +1,19 @@
...
+        return struct.pack('B', 253) + struct.pack('<H', n)
+    elif n < 0xFFFFFFFF:
```
daira: Similarly, `<= 0xFFFFFFFF`.

zc_utils.py
```diff
@@ -0,0 +1,19 @@
...
+        return struct.pack('B', 254) + struct.pack('<I', n)
+    else:
+        return struct.pack('B', 255) + struct.pack('<Q', n)
+
+assert write_compact_size(0) == b'\x00'
+assert write_compact_size(252) == b'\xFC'
+assert write_compact_size(253) == b'\xFD\xFD\x00'
+assert write_compact_size(0xFFFE) == b'\xFD\xFE\xFF'
+assert write_compact_size(0xFFFF) == b'\xFE\xFF\xFF\x00\x00'
```
daira: `b'\FD\xFF\xFF'`

zc_utils.py
```diff
@@ -0,0 +1,19 @@
...
+assert write_compact_size(0xFFFFFFFE) == b'\xFE\xFE\xFF\xFF\xFF'
+assert write_compact_size(0xFFFFFFFF) == b'\xFF\xFF\xFF\xFF\xFF\x00\x00\x00\x00'
```
daira: `b'\xFE\xFF\xFF\xFF\xFF'

Also add tests for 0x10000, 0x100000000 (8 '0's), and 0xFFFFFFFFFFFFFFFF (16 'F's).

str4d:
Rebased on master to fix merge conflicts and address @daira's comments. I also squashed the commits from the second-to-last review round.

str4d:
Fixed some issues I had when trying to use the test vectors in Rust.

I have also manually confirmed that none of the previous test vectors' Rust output is altered by this PR.

transaction.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, rand):
+        self._script = bytes([
+            rand.a(RAND_OPCODES) for i in range(rand.u8() % 10)
+        ])
+
+    def raw(self):
+        return self._script
+
+    def __bytes__(self):
+        return struct.pack('b', len(self._script)) + self._script
```
daira: Note to self: needs CompactSize.

zip_0143.py
```diff
@@ -0,0 +1,157 @@
...
+from tv_output import render_args, render_tv
+from tv_rand import Rand
+
+
+SIGHASH_ALL = 1
+SIGHASH_NONE = 2
+SIGHASH_SINGLE = 3
+SIGHASH_ANYONECANPAY = 0x80
+
+NOT_AN_INPUT = -1 # For portability of the test vectors
```
daira: I originally thought this should be ``NOT_AN_INPUT = 0xFFFFFFFF``, but in fact the [C++ code](https://github.com/zcash/zcash/blob/f0daf3915fd81fec7f0f79c4e109fdfaef44d162/src/script/interpreter.h#L23) defines it as ``UINT_MAX`` regardless of the size of ``unsigned int``, so there is no correct cross-platform value to use other than -1. As long as the JSON test vectors use -1, this will work.

PR #6, opened 2018-07-03
Additions to .gitignore; rename CRS to URS; add test vectors for Pedersen generators

daira:


sapling_generators.py
```diff
@@ -48,6 +49,9 @@ def find_group_hash(D, M):
 VALUE_COMMITMENT_VALUE_BASE = find_group_hash(b'Zcash_cv', b'v')
 VALUE_COMMITMENT_RANDOMNESS_BASE = find_group_hash(b'Zcash_cv', b'r')
 
+required_bases = 4
+PEDERSEN_BASES = [find_group_hash(b'Zcash_PH', i2leosp(32, iminus1))
+                  for iminus1 in range(0, required_bases)]
```
str4d: I'd mildly prefer that this was in `sapling_pedersen.py` using `I_D_i()`, but meh.

PR #9, opened 2018-08-22
Implement ZIP 243 test vectors.

bitcartel:


transaction.py
```diff
@@ -41,9 +62,59 @@ def __bytes__(self):
...
+        )
+
+class SpendDescription(object):
+    def __init__(self, rand):
+        self.cv = find_group_hash(b'TVRandPt', rand.b(32))
+        self.anchor = Fq(leos2ip(rand.b(32)))
+        self.nullifier = rand.b(32)
+        self.rk = rand.b(32)
+        self.proof = GrothProof(rand)
+        self.spendAuthSig = rand.b(64) # Invalid
```
arielgabizon: what does the invalid comment mean exactly?

ebfull: It's just indicating that it's invalid (a random blob) because the purpose is to test the serialization and SignatureHash behavior, not components like signatures or proofs.

arielgabizon: right but that seems true of all the fields, so I didn't understand what was special about this line.

zip_0243.py
```diff
@@ -0,0 +1,175 @@
...
+        digest.update(bytes(desc.proof))
+    return digest.digest()
+
+def getHashShieldedOutputs(tx):
+    digest = blake2b(digest_size=32, person=b'ZcashSOutputHash')
+    for desc in tx.vShieldedOutputs:
+        digest.update(bytes(desc))
+    return digest.digest()
+
+def signature_hash(scriptCode, tx, nIn, nHashType, amount, consensusBranchId):
```
arielgabizon: Wonder if there's a way to do this pr with less code duplication - only the two lines about hashes of spends and outputs are new.

bitcartel: A good idea to refactor this and reduce code duplication.  Something for post-Sapling activation.

transaction.py
bitcartel: You make a fair point that all the data is test data.  Its just a human friendly reminder that a computation or derivation has not taken place, that the field is junk.  To be consistent with existing fields labelled this way, this PR sticks with signature and proof.

PR #10, opened 2018-10-11
Update ZIP 243 test vectors with valid rk values

str4d:


sapling_jubjub.py
```diff
@@ -142,6 +142,14 @@ def __str__(self):
...
 
 class Point(object):
+    @staticmethod
+    def rand(rand):
+        while True:
+            data = rand.b(32)
+            p = Point.from_bytes(data)
+            if p:
+                return p
+
```
daira: For clarity, ``if not u:`` in the ``from_bytes`` method on line 167 should be ``if u is not None:``. (It's technically equivalent because a not-None ``Fq`` object is always truthy even if it represents zero. However, that took me about 30 seconds to think about.)

sapling_jubjub.py
```diff
@@ -142,6 +142,14 @@ def __str__(self):
 JUBJUB_COFACTOR = Fr(8)
 
 class Point(object):
+    @staticmethod
+    def rand(rand):
+        while True:
+            data = rand.b(32)
+            p = Point.from_bytes(data)
+            if p:
```
daira: Change this to ``if p is not None:``.

sapling_jubjub.py
str4d: I assume you mean that it should be `if u is None:`

str4d:
Addressed @daira's comments, and fixed an issue introduced in #9 that changed the ZIP 143 test vectors.

PR #13, opened 2020-07-02
limit the amount note_v to max_money

paracetamolo:
After the addition of Amount to librustzcash, we enforced it everywhere in our codebase (Sapling integration in Tezos). 
We recently generated some vectors with these scripts and they were failing, hence this MR.


str4d:
We won't be enforcing this in these test vectors, because the Sapling protocol is defined on (and must be correct for) 64-bit values. `MAX_MONEY` is only enforced at the `valueBalance` boundary.

PR #17, opened 2021-03-30
Add iso-Pallas, SWU hash-to-curve, and Sinsemilla

defuse:


defuse:
I tried reducing the duplication between Pallas and iso-Pallas using inheritance or ownership but it turned out to be too error-prone, i.e. hard to be sure that neither curve wasn't accidentally using the operations of the other, so I left the duplication in.

orchard_iso_pallas.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+Point.ZERO = Point.identity()
+
+x = Fp(2)
+y2 = x * x * x + PALLAS_ISO_A * x + PALLAS_ISO_B
+y = y2.sqrt()
+assert y is not None
+
+Point.GENERATOR = Point(x, y)
```
defuse, line 168: Is this actually a generator? Is iso-Pallas also prime-order?

daira: Yes. Isogenous curves always have the same order.

orchard_iso_pallas.py
```diff
@@ -0,0 +1,176 @@
...
+
+        x3 = x * x * x
+        y2 = x3 + PALLAS_ISO_A * x + PALLAS_ISO_B
+
+        y = y2.sqrt()
+        if y is None:
+            return None
+
+        if y.s % 2 != y_sign:
+            y = Fp.ZERO - y
```
daira, line 38:
```suggestion
            y = -y
```

orchard_pallas.py
```diff
@@ -185,6 +194,10 @@ def double(self):
         x = λ*λ - self.x - self.x
         y = λ*(self.x - x) - self.y
         return Point(x, y)
+    
+    def extract(self):
+        assert not self.is_identity
```
daira, line 199:
```suggestion
        if self.is_identity:
            return 0
```

orchard_sinsemilla.py
```diff
@@ -0,0 +1,175 @@
...
+    b1_ctx = blake2b(digest_size=64, person=i2beosp(128,0))
+    b1_ctx.update(b[0] + i2beosp(8, 1) + dst_prime)
+    b.append(b1_ctx.digest())
+    assert len(b[1]) == b_in_bytes
+
+    for i in range(2, ell + 1):
+        bi_input = b"\x00" * b_in_bytes
+
+        for j in range(0, i):
+            bi_input = sxor(bi_input, b[j])
```
daira, line 50:
```suggestion
        bi_input = sxor(b[0], b[i-1])
```

orchard_sinsemilla.py
```diff
@@ -0,0 +1,175 @@
...
+
+        bi_input += i2beosp(8, i) + dst_prime
+
+        bi_ctx = blake2b(digest_size=64, person=i2beosp(128,0))
+        bi_ctx.update(bi_input)
+
+        b.append(bi_ctx.digest())
+        assert len(b[i]) == b_in_bytes
+
+    return b''.join(b)[0:len_in_bytes]
```
daira, line 62:
```suggestion
    return b''.join(b[1:])[0:len_in_bytes]
```

orchard_sinsemilla.py
```diff
@@ -0,0 +1,175 @@
...
+    uniform_bytes = expand_message_xmd(msg, dst, len_in_bytes)
+
+    elements = []
+    for i in range(0, count):
+        for j in range(0, m):
+            elm_offset = L * (j + i * m)
+            tv = uniform_bytes[elm_offset:elm_offset+L]
+            elements.append(Fp(beos2ip(tv), False))
+    
+    assert len(elements) == 2
```
daira, line 82:
```suggestion
    assert len(elements) == count
```

orchard_sinsemilla.py
```diff
@@ -0,0 +1,175 @@
...
+    for m_i in m:
+        acc = acc + group_hash(b"z.cash:SinsemillaS", i2leosp(32, m_i)) + acc
+    
+    return acc
+
+def sinsemilla_hash(d, m):
+    return sinsemilla_hash_to_point(d, m).extract()
+
+# m_bytes MUST be a b"byte string", otherwise it could be parsed as hex!
+def sinsemilla_hash_bytes(d, m_bytes):
```
daira, line 170:
```suggestion
def sinsemilla_hash_bytes(d, m_bytes):
    assert isinstance(m_bytes, bytes)
```

orchard_sinsemilla.py
```diff
@@ -0,0 +1,175 @@
...
+    return sinsemilla_hash(d, BitArray(m_bytes))
+
+if __name__ == "__main__":
+    sh = sinsemilla_hash_bytes(b"whatever", b"whatever2")
```
daira, line 174:
```suggestion
    sh = sinsemilla_hash_bytes(b"z.cash:test", b"Trans rights now!")
```

orchard_iso_pallas.py
```diff
@@ -0,0 +1,176 @@
...
+    def __init__(self, x, y, is_identity=False):
+        self.x = x
+        self.y = y
+        self.is_identity = is_identity
+
+        if is_identity:
+            assert self.x == Fp.ZERO
+            assert self.y == Fp.ZERO
+        else:
+            assert self.y * self.y == self.x * self.x * self.x + PALLAS_ISO_A * self.x + PALLAS_ISO_B
```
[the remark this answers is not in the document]
daira: Yes.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+    return sinsemilla_hash_to_point(d, m).extract()
+
+def sinsemilla_hash_bytes(d, m_bytes):
+    assert isinstance(m_bytes, bytes)
+    return sinsemilla_hash(d, BitArray(m_bytes))
+
+if __name__ == "__main__":
+    # This is the Pallas test vector from the Sage and Rust code (in affine coordinates).
+    gh = group_hash(b"z.cash:test", b"Trans rights now!")
+    assert gh == Point(Fp(10899331951394555178876036573383466686793225972744812919361819919497009261523),
```
dconnolly, line 176: 🎉

dconnolly:
Can we use `render_*` in `__main__` to make it easy to get the Rust/zcash/json renderings of these values in a reproducible way? 🙏

orchard_pallas.py
```diff
@@ -185,6 +194,11 @@ def double(self):
         x = λ*λ - self.x - self.x
         y = λ*(self.x - x) - self.y
         return Point(x, y)
+    
+    def extract(self):
+        if self.is_identity:
+            return 0
```
defuse, line 200:
```suggestion
            return Fp.ZERO
```

defuse:
ACK @daira's changes aside from the Fp.ZERO suggested change

str4d:
> 
> 
> Can we use `render_*` in `__main__` to make it easy to get the Rust/zcash/json renderings of these values in a reproducible way? 🙏

Agreed; we shouldn't be doing any non-trivial assertions in `__main__` here, since this _is_ where we generate the test vectors.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+
+import math
+
+import orchard_iso_pallas
+
+from pyblake2 import blake2b, blake2s
```
daira, line 8:
```suggestion
from pyblake2 import blake2b
```

daira (dismissed):
ACK

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+
+def sinsemilla_hash_to_point(d, m):
+    assert isinstance(m, BitArray)
+    n = cldiv(m.len, SINSEMILLA_K)
+    m = pad(n, m)
+    acc = group_hash(b"z.cash:SinsemillaQ", d)
+    #print("acc", acc)
+
+    for m_i in m:
+        acc = acc + group_hash(b"z.cash:SinsemillaS", i2leosp(32, m_i)) + acc
```
str4d, line 161: This is using the complete addition implemented on `pallas.Point`. It needs to instead use incomplete addition, which maps all edge cases to ⊥ instead of their complete outputs.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+import math
+
+import orchard_iso_pallas
+
+from pyblake2 import blake2b, blake2s
+from orchard_pallas import Fp, p, q, PALLAS_B, Point
+from orchard_iso_pallas import PALLAS_ISO_B, PALLAS_ISO_A
+from sapling_utils import i2beosp, cldiv, beos2ip, i2leosp, lebs2ip
+from binascii import hexlify
+from bitstring import BitArray
```
str4d, line 13: AFAICT this is an external module, so we need to update the requirements in the README.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+        #print("acc", acc)
+
+    return acc
+
+def sinsemilla_hash(d, m):
+    return sinsemilla_hash_to_point(d, m).extract()
+
+def sinsemilla_hash_bytes(d, m_bytes):
+    assert isinstance(m_bytes, bytes)
+    return sinsemilla_hash(d, BitArray(m_bytes))
```
str4d, line 171: The `BitArray` documentation does not specify the endianness with which it returns the bits in each byte. This is not precise enough for our usage, so I would prefer finding an alternative bit iterator. And in any case, we don't need `sinsemilla_hash_bytes` for our use cases (we don't have any inputs that are an exact number of bytes; for example, Commit<sup>ivk</sup> takes 510 bits of input).

str4d (changes requested):
Reviewed with @therealyingtong.

orchard_iso_pallas.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+Point.ZERO = Point.identity()
+
+x = Fp(2)
+y2 = x * x * x + PALLAS_ISO_A * x + PALLAS_ISO_B
+y = y2.sqrt()
+assert y is not None
+
+Point.GENERATOR = Point(x, y)
```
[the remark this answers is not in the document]
daira: I also wondered this. Technically any non-zero point will do, though.

If we were to "officially" define a generator, it would be iso_map<sup>-1</sup>((-1, 2)).

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+        bi_ctx.update(bi_input)
+
+        b.append(bi_ctx.digest())
+        assert len(b[i]) == b_in_bytes
+
+    return b''.join(b[1:])[0:len_in_bytes]
+
+def hash_to_field(msg, dst):
+    k = 256
+    count = 2
```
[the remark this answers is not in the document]
daira: I think `dst` being a parameter in the draft is a bit weird. It needs to be an input. And `count` is always 2 when used for `hash_to_curve`; the draft only needs it to be a parameter in order to support non-random-oracle encoding, which we don't use. This code looks fine to me.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+    m = 1
+
+    L = cldiv(math.ceil(math.log2(p)) + k, 8)
+    assert L == 512/8
+
+    len_in_bytes = count * 1 * L
```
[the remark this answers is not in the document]
daira: I don't understand what m is here. There's no m in the definition of [expand_message_xmd](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-10.html#name-expand_message_xmd-2).

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+    uniform_bytes = expand_message_xmd(msg, dst, len_in_bytes)
+
+    elements = []
+    for i in range(0, count):
+        for j in range(0, m):
+            elm_offset = L * (j + i * m)
+            tv = uniform_bytes[elm_offset:elm_offset+L]
+            elements.append(Fp(beos2ip(tv), False))
```
daira, line 77:
```suggestion
    L = cldiv(math.ceil(math.log2(p)) + k, 8)
    assert L == 512/8

    len_in_bytes = count * L
    uniform_bytes = expand_message_xmd(msg, dst, len_in_bytes)

    elements = []
    for i in range(0, count):
        elm_offset = L * i
        tv = uniform_bytes[elm_offset:elm_offset+L]
        elements.append(Fp(beos2ip(tv), False))
```

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+
+    return elements
+
+def map_to_curve_simple_swu(u):
+    zero = Fp(0)
+    assert zero.inv() == Fp(0)
+
+    A = PALLAS_ISO_A
+    B = PALLAS_ISO_B
+    Z = Fp(-13, False)
```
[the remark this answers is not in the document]
daira: Z was calculated like this: https://github.com/zcash/pasta/blob/d10932faf06278510b59093a20fe225ddb9ea402/hashtocurve.sage#L28-L59

orchard_sinsemilla.py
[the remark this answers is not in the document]
str4d: It's the size of the extension field, which is 1 for us.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+    x1 = x1.inv()
+    e1 = x1 == Fp(0)
+    x1 = x1 + Fp(1)
+
+    if e1:
+        x1 = c2
+    else:
+        x1 = x1
+
+    x1 = x1 * c1      # x1 = (-B / A) * (1 + (1 / (Z^2 * u^4 + Z * u^2)))
```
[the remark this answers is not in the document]
daira: I have no opinion on this. I think it's reasonable to follow either naming.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+    gx1 = x1.exp(2)
+    gx1 = gx1 + A
+    gx1 = gx1 * x1
+    gx1 = gx1 + B             # gx1 = g(x1) = x1^3 + A * x1 + B
+    x2 = tv1 * x1            # x2 = Z * u^2 * x1
+    tv2 = tv1 * tv2
+    gx2 = gx1 * tv2           # gx2 = (Z * u^2)^3 * gx1
```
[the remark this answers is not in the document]
daira: The comment "gx2 = (Z * u^2)^3 * gx1" in the I-D indeed looks completely wrong. gx2 is supposed to be g(x2) = g(Z.u^2.x1) = Z^3.u^6.x1^3 + Z^2.u.x1 + B, which is not the same as Z^3.u^6.(x1^3 + A.x1 + B). I will raise this on the cfrg list.

Aside: I find the reuse of variable names in the I-D to be completely horrible. It might be trying to indicate how many variables are needed, but that kind of low-level optimization doesn't really belong there.

orchard_sinsemilla.py
daira: It looks like this dependency can just be removed in favour of using strings of '0' and '1'.

daira (commented):
Confirmed @str4d's comments.

orchard_sinsemilla.py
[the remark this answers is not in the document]
daira: Hmm, I don't understand why the `group_hash` test vector is passing. It should exercise both the x1 and x2 cases, and the Sage and Rust implementations don't have this bug. 😕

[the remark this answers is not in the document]
daira: Applying this patch still passes. The plot thickens.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+    y = y2.sqrt()
+
+    e3 = u.sgn0() == y.sgn0()
+
+    y = y if e3 else -y  #y = CMOV(-y, y, e3)
+
+    return orchard_iso_pallas.Point(x, y)
+
+
+def group_hash(d, m):
```
therealyingtong, line 128: nit: `group_hash` is used elsewhere in the Orchard protocol besides Sinsemilla. we could move these into another module.

str4d: I'm fine with this living here, since it is a simple wrapper around Sinsemilla.

str4d: Wait, sorry, I had this the wrong way round in my head. Yes, this should be in a separate file (e.g. `orchard_group_hash.py`).

orchard_sinsemilla.py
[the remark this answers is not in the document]
daira: OK, I've checked that the two expressions are equivalent given the choice of x1 (and so Appendix F.1 is correct). No need to change this.

orchard_iso_pallas.py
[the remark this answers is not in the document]
defuse: No reason at all, I was just trying to come up with any nonzero point on the curve.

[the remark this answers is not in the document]
str4d: Then we should document that this is not a formally-selected common generator for Pallas, or else everyone will assume it is.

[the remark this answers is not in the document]
defuse: Documented, thanks!

defuse:
Note to self: figure out why the missing division by Z didn't lead to a test vector mismatch.

orchard_sinsemilla.py
defuse: Good catch, I'll delete this code

orchard_sinsemilla.py
defuse: Is there a reason to use incomplete addition formulas in the test vector implementation, since returning⊥ implies a discrete log relation has been found? My thinking is that for test vectors we want the most obviously-correct output, so perhaps it would be better to keep using the existing `+` but assert that the code isn't hitting any of the 'incomplete' cases?

orchard_sinsemilla.py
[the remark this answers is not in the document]
defuse: Note to self: double check this when I have time.

orchard_sinsemilla.py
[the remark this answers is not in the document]
defuse: I'm interpreting this as the PR code doesn't need to be changed (rather than @str4d's patch doesn't need to be changed), please correct me if I'm wrong.

orchard_sinsemilla.py
```diff
@@ -0,0 +1,182 @@
...
+    tv1 = Z * u.exp(2)
+    tv2 = tv1.exp(2)
+    x1 = tv1 + tv2
+
+    x1 = x1.inv()
+    e1 = x1 == Fp(0)
+    x1 = x1 + Fp(1)
+
+    if e1:
+        x1 = c2
```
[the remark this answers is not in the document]
defuse: This should be fixed now after I applied @str4d's patch to set `c2 = Fp(-1) / Z` (which matches Appendex F.2 now), please correct me if I'm wrong.

defuse:
All of the output-affecting comments have been addressed (I think), I will come back to this on Monday to finish up the refactor-y-type things.

orchard_sinsemilla.py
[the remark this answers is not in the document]
str4d: That was indeed the fix I was proposing for this.

orchard_sinsemilla.py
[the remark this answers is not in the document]
str4d: Correct; the PR is fine as-is if these two are equivalent. I personally would prefer following the ID specification here rather than Appendix F.2 but I'm not going to block on it.

orchard_sinsemilla.py
str4d: I'd be fine with asserting we don't hit any edge cases instead.

daira:
> Note to self: figure out why the missing division by Z didn't lead to a test vector mismatch.

That will be because we don't have a test vector for map_to_curve with u = 0. I'll add one.

orchard_pallas.py
```diff
@@ -182,6 +182,13 @@ def __add__(self, a):
             else:
                 return self.double()
 
+    def checked_incomplete_add(self, a):
+        assert self != a
+        assert self != -a
+        assert self != Point.identity()
+        assert self != Point.identity()
```
str4d, line 189:
```suggestion
        assert a != Point.identity()
```

str4d (changes requested):
utACK once the bug below is fixed.

str4d (approved):
utACK

orchard_group_hash.py
```diff
@@ -0,0 +1,167 @@
...
+    x1 = tv1 + tv2
+
+    x1 = x1.inv()
+    e1 = x1 == Fp(0)
+    x1 = x1 + Fp(1)
+
+    if e1:
+        x1 = c2
+    else:
+        x1 = x1
```
daira, line 104:
```suggestion
    x1 = c2 if e1 else x1    # If (tv1 + tv2) == 0, set x1 = -1 / Z
```

orchard_group_hash.py
```diff
@@ -0,0 +1,167 @@
...
+
+    x = x1 if e2 else x2    # If is_square(gx1), x = x1, else x = x2
+    yy = gx1 if e2 else gx2  # If is_square(gx1), yy = gx1, else yy = gx2
+    y = yy.sqrt()
+
+    e3 = u.sgn0() == y.sgn0()
+
+    y = y if e3 else -y  #y = CMOV(-y, y, e3)
+
+    return orchard_iso_pallas.Point(x, y).iso_map()
```
daira, line 125: The algorithm in Appendix F.2 does not include isogeny map in its `map_to_curve_simple_swu`. Consider refactoring so that this more closely matches the draft.

orchard_iso_pallas.py
```diff
@@ -0,0 +1,178 @@
...
+        else:
+            return Point(Fp(self.x.s), -Fp(self.y.s))
+
+    def __add__(self, a):
+        if self.is_identity:
+            return a
+        elif a.is_identity:
+            return self
+        else:
+            # <https://core.ac.uk/download/pdf/10898289.pdf> section 4.1
```
daira, line 100:
```suggestion
            # Hüseyin Hışıl. “Elliptic Curves, Group Law, and Efficient Computation”. PhD thesis.
            # <https://core.ac.uk/download/pdf/10898289.pdf> section 4.1
```

orchard_iso_pallas.py
```diff
@@ -0,0 +1,178 @@
...
+                return Point(x3, y3)
+
+    def __sub__(self, a):
+        return (-a) + self
+
+    def double(self):
+        if self.is_identity:
+            return self
+
+        # <https://core.ac.uk/download/pdf/10898289.pdf> section 4.1
```
daira, line 122:
```suggestion
        # Hüseyin Hışıl. “Elliptic Curves, Group Law, and Efficient Computation”. PhD thesis.
        # <https://core.ac.uk/download/pdf/10898289.pdf> section 4.1
```

orchard_iso_pallas.py
```diff
@@ -0,0 +1,178 @@
...
+            return bytes([0] * 32)
+
+        buf = bytes(self.x)
+        if self.y.s % 2 == 1:
+            buf = buf[:31] + bytes([buf[31] | (1 << 7)])
+        return buf
+
+    def __eq__(self, a):
+        if a is None:
+            return False
```
daira, line 148: Why is this needed (rather than just failing on `a.is_identity`)?

daira (approved):
utACK with minor comments.

I added three commits for direct testing of `map_to_curve_simple_swu`; these should also be reviewed.

orchard_group_hash.py
daira: This is done in my first commit (322aff1777f8ccd74bbea85eb05d65d845a5827e).

orchard_group_hash.py
```diff
@@ -136,21 +136,17 @@ def group_hash(d, m):
...
-    test_vectors = [{
-        'domain': domain,
-        'msg': msg,
-        'point': bytes(group_hash(domain, msg)),
-    } for (domain, msg) in test_vectors]
+    for (domain, msg, point) in group_hash_test_vectors:
+        gh = group_hash(domain, msg)
+        assert gh == point
 
+    print("group_hash (Pallas):")
```
str4d, line 178: This is not how this test vector renderer is supposed to be used; the output is intended to be entirely copy-pasteable, and untyped text output interferes with that.

orchard_group_hash.py
```diff
@@ -136,21 +136,17 @@ def group_hash(d, m):
 
 
 def main():
-    test_vectors = [
-        (b"z.cash:test", b"Trans rights now!"),
+    group_hash_test_vectors = [
+        # This is the Pallas test vector from the Sage and Rust code (in affine coordinates).
+        (b"z.cash:test", b"Trans rights now!", Point(Fp(10899331951394555178876036573383466686793225972744812919361819919497009261523),
+                                                     Fp(851679174277466283220362715537906858808436854303373129825287392516025427980))),
```
str4d, line 171: This refactor interferes with the test vector generator. We are supposed to be _generating_ test vectors here, not checking them. I special-cased the one from the Sage code, but the intention is that we'll use randomness to generate test vectors for which this kind of check would be pointlessly tautological.

orchard_group_hash.py
```diff
@@ -136,6 +136,35 @@ def group_hash(d, m):
...
+        (Fp(0x123456789abcdef123456789abcdef123456789abcdef123456789abcdef0123),
+                orchard_iso_pallas.Point(Fp(16711718778908753690082328243251803703269853000652055785581237369882690082595),
+                                         Fp(1764705856161931038824461929646873031992914829456409784642560948827969833589))),
+    ]
+
+    for (u, point) in map_to_curve_test_vectors:
+        P = map_to_curve_simple_swu(u)
+        assert P == point
+
+    print("map_to_curve_simple_swu (Pallas):")
```
str4d, line 153: These should be separately-generated outputs, not in-line the same, as this prevents the output being directly copy-pasteable or piped to a file.

str4d (changes requested):

orchard_group_hash.py
```diff
@@ -0,0 +1,168 @@
...
+        'orchard_group_hash',
+        (
+            ('domain', 'Vec<u8>'),
+            ('msg', 'Vec<u8>'),
+            ('point', '[u8; 32]'),
+        ),
+        [{
+            'domain': domain,
+            'msg': msg,
+            'point': bytes(group_hash(domain, msg)),
```
daira, line 162: Why recompute `group_hash` here when we already checked it is equal to the point in `fixed_test_vectors`?

orchard_map_to_curve.py
```diff
@@ -0,0 +1,41 @@
...
+    render_tv(
+        render_args(),
+        'orchard_map_to_curve',
+        (
+            ('u', '[u8; 32]'),
+            ('point', '[u8; 32]'),
+        ),
+        [{
+            'u': bytes(u),
+            'point': bytes(map_to_curve_simple_swu(u)),
```
daira, line 35: Similarly, why recompute `map_to_curve_simple_swu` here?

daira (approved):
utACK with comments.

orchard_group_hash.py
daira: Never mind, I see why now that you've added the random generation code.

orchard_iso_pallas.py
daira:
```suggestion
```

str4d (approved):
ACK f8008a01fdc76f384cf62e3283ea775281070b82

orchard_iso_pallas.py
str4d: Failing on `a.is_identity` would crash the generator (as `None` doesn't have that method). It's possible to get `None` as an output of `Point.from_bytes`, so it needs to be handled in `__eq__`.

str4d:
I'm happy that the few remaining open comment threads have resolutions.

orchard_group_hash.py
```diff
@@ -0,0 +1,185 @@
...
+        } for (domain, msg) in test_vectors],
+    )
```
dconnolly, line 181: Yay! Thank you 🙏

PR #28, opened 2021-05-03
Add Orchard commitments

therealyingtong:
Commitment schemes used in Orchard https://zips.z.cash/protocol/nu5.pdf#concretecommit

`HomomorphicPedersenCommit` -> `ValueCommit`
`SinsemillaCommit` -> `NoteCommit`
`SinsemillaShortCommit` -> `CommitIvk`


orchard_pallas.py
```diff
@@ -31,6 +31,13 @@ class Fp(FieldElement):
     def from_bytes(buf):
         return Fp(leos2ip(buf), strict=True)
 
+    def random(randbytes):
+        while True:
+            try:
+                return Fp(leos2ip(randbytes(32)), strict=True)
```
str4d, line 37: Use the `Rand` type from `tv_rand.py` instead:
```suggestion
                return Fp(leos2ip(rand.b(32)), strict=True)
```

orchard_commitments.py
```diff
@@ -0,0 +1,74 @@
...
+def rcm_trapdoor(randbytes):
+    return Scalar.random(randbytes)
+
+# https://zips.z.cash/protocol/nu5.pdf#concreteorchardnotecommit
+def commit_ivk(rivk: Scalar, ak: Fp, nk: Fp):
+    return Scalar(sinsemilla_short_commit(
+        rivk,
+        b"z.cash: Orchard-CommitIvk",
+        i2lebsp(L_ORCHARD_BASE, ak.s) + i2lebsp(L_ORCHARD_BASE, nk.s)
+    ).s)
```
str4d, line 51: Hmm, in the spec, Commit<sup>ivk</sup>.Output is [defined as](https://zips.z.cash/protocol/nu5.pdf#abstractcommit) a scalar mod q<sub>P</sub>, not r<sub>P</sub>. And IVK parsing [MUST reject if the result is not in F<sub>q<sub>P</sub></sub>](https://zips.z.cash/protocol/nu5.pdf#orchardinviewingkeyencoding). So I think `commit_ivk` should not convert from a `Base` to a `Scalar` here:
```suggestion
    return sinsemilla_short_commit(
        rivk,
        b"z.cash: Orchard-CommitIvk",
        i2lebsp(L_ORCHARD_BASE, ak.s) + i2lebsp(L_ORCHARD_BASE, nk.s)
    )
```

The [Orchard key components spec](https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents) mentions that the output of Commit<sup>ivk</sup> is used as a KA<sup>Orchard</sup> private key `ivk`, so it's unclear there whether `ivk` is supposed to be typed as `Base` or `Scalar`, but given the above I think it should be typed `Base` and then explicitly converted to a `Scalar` at the point we use it. I'll make a similar change to the `orchard` crate.

cc @daira for verification.

orchard_commitments.py
```diff
@@ -0,0 +1,74 @@
...
+
+# https://zips.z.cash/protocol/nu5.pdf#concretehomomorphiccommit
+def homomorphic_pedersen_commitment(rcv: Scalar, D, v: Scalar):
+    return group_hash(D, b"v") * v + group_hash(D, b"r") * rcv
+
+def value_commit(rcv: Scalar, v: Scalar):
+    return homomorphic_pedersen_commitment(rcv, b"z.cash:Orchard-cv", v)
+
+def rcv_trapdoor(randbytes):
+    return Scalar.random(randbytes)
```
str4d, line 22:
```suggestion
def rcv_trapdoor(rand):
    return Scalar.random(rand)
```

orchard_commitments.py
```diff
@@ -0,0 +1,74 @@
...
+# https://zips.z.cash/protocol/nu5.pdf#concreteorchardnotecommit
+def note_commit(rcm, g_d, pk_d, v, rho, psi):
+    return sinsemilla_commit(
+        rcm,
+        b"z.cash: Orchard-NoteCommit",
+        g_d + pk_d + i2lebsp(64, v) + i2lebsp(L_ORCHARD_BASE, rho.s) + i2lebsp(L_ORCHARD_BASE, psi.s)
+    )
+
+def rcm_trapdoor(randbytes):
+    return Scalar.random(randbytes)
```
str4d, line 43:
```suggestion
def rcm_trapdoor(rand):
    return Scalar.random(rand)
```

orchard_commitments.py
```diff
@@ -0,0 +1,74 @@
...
+# https://zips.z.cash/protocol/nu5.pdf#concreteorchardnotecommit
+def commit_ivk(rivk: Scalar, ak: Fp, nk: Fp):
+    return Scalar(sinsemilla_short_commit(
+        rivk,
+        b"z.cash: Orchard-CommitIvk",
+        i2lebsp(L_ORCHARD_BASE, ak.s) + i2lebsp(L_ORCHARD_BASE, nk.s)
+    ).s)
+
+def rivk_trapdoor(randbytes):
+    return Scalar.random(randbytes)
```
str4d, line 54:
```suggestion
def rivk_trapdoor(rand):
    return Scalar.random(rand)
```

orchard_commitments.py
```diff
@@ -0,0 +1,74 @@
...
+def test_value_commit():
+    from random import Random
+    from orchard_generators import VALUE_COMMITMENT_RANDOMNESS_BASE, VALUE_COMMITMENT_VALUE_BASE
+
+    rng = Random(0xabad533d)
+    def randbytes(l):
+        ret = []
+        while len(ret) < l:
+            ret.append(rng.randrange(0, 256))
+        return bytes(ret)
```
str4d, line 66:
```suggestion
        return bytes(ret)
    rand = Rand(randbytes)
```

orchard_commitments.py
```diff
@@ -0,0 +1,74 @@
...
+
+    rcv = rcv_trapdoor(randbytes)
```
str4d, line 68:
```suggestion
    rcv = rcv_trapdoor(rand)
```

str4d (changes requested):

orchard_commitments.py
daira: The protocol spec types ivk as an integer in [1, q_P).

str4d: Yep, and that is enforced by construction: `SinsemillaCommit` never returns the identity (which we assert in the test vector generator for Sinsemilla), and thus the zero-output case of ExtractP is never triggered.

So my comment stands: change `ivk` to be typed as `Base` and then explicitly convert to `Scalar` when necessary.

orchard_commitments.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+def sinsemilla_short_commit(r: Scalar, D, M):
+    return sinsemilla_commit(r, D, M).extract()
+
+# https://zips.z.cash/protocol/nu5.pdf#concreteorchardnotecommit
+def note_commit(rcm, g_d, pk_d, v, rho, psi):
+    return sinsemilla_commit(
+        rcm,
+        b"z.cash: Orchard-NoteCommit",
```
str4d, line 38:
```suggestion
        b"z.cash:Orchard-NoteCommit",
```

orchard_commitments.py
```diff
@@ -0,0 +1,76 @@
...
+    )
+
+def rcm_trapdoor(rand):
+    return Scalar.random(rand)
+
+# https://zips.z.cash/protocol/nu5.pdf#concreteorchardnotecommit
+def commit_ivk(rivk: Scalar, ak: Fp, nk: Fp):
+    return sinsemilla_short_commit(
+        rivk,
+        b"z.cash: Orchard-CommitIvk",
```
str4d, line 49:
```suggestion
        b"z.cash:Orchard-CommitIvk",
```

str4d (changes requested):

str4d (approved):

nuttycom (approved):
utACK

daira (commented):
Post-hoc ACK for the last three commits.

PR #29, opened 2021-05-06
Orchard key components

### LINKED ISSUE #24: Implement Orchard key components test vectors
Equivalent of `sapling_key_components.py`; this should:

- Generate a random Orchard spending key.
- Derive all of the various key components from it.
- Generate a note that is spendable by the spending key.
- Derive the note's commitment and nullifier.

daira:
Closes #24.

### DIFF
--- a/orchard_key_components.py
+++ b/orchard_key_components.py
@@ -43,7 +43,7 @@ def prf_nf_orchard(nk, rho):
 
 def derive_nullifier(nk, rho: Fp, psi: Fp, cm):
     scalar = prf_nf_orchard(nk, rho) + psi  # addition mod p
-    point = NULLIFIER_K_BASE * to_scalar(scalar) + cm
+    point = NULLIFIER_K_BASE * Scalar(scalar.s) + cm
     return point.extract()
 
 #
--- a/orchard_key_components.py
+++ b/orchard_key_components.py
@@ -42,7 +42,7 @@ def prf_nf_orchard(nk, rho):
     return poseidon_hash(nk, rho)
 
 def derive_nullifier(nk, rho: Fp, psi: Fp, cm):
-    scalar = to_base(prf_nf_orchard(nk, rho)) + psi  # addition mod p
+    scalar = prf_nf_orchard(nk, rho) + psi  # addition mod p
     point = NULLIFIER_K_BASE * to_scalar(scalar) + cm
     return point.extract()
 
--- a/orchard_key_components.py
+++ b/orchard_key_components.py
@@ -1,14 +1,13 @@
 #!/usr/bin/env python3
 import sys; assert sys.version_info[0] >= 3, "Python 3 required."
 
-from pyblake2 import blake2b, blake2s
+from ff1 import ff1_aes256_encrypt
 
 from orchard_generators import NULLIFIER_K_BASE, SPENDING_KEY_BASE, group_hash
 from orchard_pallas import Fp, Scalar, Point
 from orchard_poseidon_hash import poseidon_hash
-from orchard_merkle_tree import MERKLE_DEPTH
 from orchard_commitments import commit_ivk, note_commit
-from utils import leos2bsp, leos2ip, i2leosp
+from utils import leos2bsp, leos2ip, i2leosp, i2lebsp, lebs2osp
 from tv_output import render_args, render_tv
 
 #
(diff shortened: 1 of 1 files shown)


orchard_key_components.py
```diff
@@ -37,6 +38,14 @@ def diversify_hash(d):
         P = group_hash(b'z.cash:Orchard-gd', b'')
     return P
 
+def prf_nf_orchard(nk, rho):
+    return poseidon_hash(nk, rho)
+
+def derive_nullifier(nk, rho: Fp, psi: Fp, cm):
+    scalar = to_base(prf_nf_orchard(nk, rho)) + psi  # addition mod p
```
daira, line 45: `poseidon_hash` already returns `Fp`, I think, in which case the `to_base` shouldn't be necessary.

orchard_key_components.py
```diff
@@ -37,6 +38,14 @@ def diversify_hash(d):
...
+    point = NULLIFIER_K_BASE * to_scalar(scalar) + cm
```
daira, line 46: Similarly, `scalar` is not a byte array so we shouldn't call `leos2ip` on it, as `to_scalar` does. It should be something like `Scalar(scalar.s)` I think.

daira (commented):
utACK

orchard_key_components.py
```diff
@@ -0,0 +1,153 @@
...
+        if self.ask == Scalar.ZERO:
+            raise ValueError("invalid spending key")
+
+        self.akP = SPENDING_KEY_BASE * self.ask
+        if bytes(self.akP)[-1] & 0x80 != 0:
+            self.ask = -self.ask
+
+        self.ak = self.akP.extract()
+        if commit_ivk(self.rivk, self.ak, self.nk) is None:
+            raise ValueError("invalid spending key")
```
str4d, line 69: Because `sinsemilla_short_commit` uses `checked_incomplete_add` which asserts that exceptional cases do not occur, we will never hit `None` here:
```suggestion
        assert commit_ivk(self.rivk, self.ak, self.nk) is not None
```

orchard_key_components.py
```diff
@@ -0,0 +1,153 @@
...
+        self.ovk = R[32:]
+
+    def ivk(self):
+        return commit_ivk(self.rivk, self.ak, self.nk)
+
+    def ovk(self):
+        return prf_expand(self.data, b'\x02')[:32]
+
+    def default_d(self):
+        return i2leosp(88, 1337)
```
therealyingtong, line 87: From protocol spec: "The diversified payment address with diversifier index 0 is called the default diversified payment address."

```suggestion
    def default_d(self):
        index = i2lebsp(88, 0)
        return lebs2osp(ff1_aes256_encrypt(self.dk(), index))
```

orchard_key_components.py
```diff
@@ -0,0 +1,153 @@
...
+        return diversify_hash(self.default_d()) * Scalar(self.ivk().s)
+
+
+def main():
+    args = render_args()
+
+    test_vectors = []
+    for i in range(0, 10):
+        sys.stdout.write(".")
+        sys.stdout.flush()
```
str4d, line 99: These will interfere with the output:
```suggestion
```

orchard_key_components.py
```diff
@@ -0,0 +1,153 @@
...
+        sk = SpendingKey(bytes([i] * 32))
+        fvk = FullViewingKey(sk)
+        note_v = (2548793025584392057432895043257984320*i) % 2**64
+        note_r = Scalar(8890123457840276890326754358439057438290574382905).exp(i+1)
+        note_rho = Fp(342358729643275392567239275209835729829*i)
+        note_psi = Fp(432592604358294371936572103719358723958*i)
```
therealyingtong, line 105: We can generate randomness using `Fp.random(rand)`.

orchard_key_components.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def default_d(self):
+        return i2leosp(88, 1337)
+
+    def default_pkd(self):
+        return diversify_hash(self.default_d()) * Scalar(self.ivk().s)
+
+
+def main():
+    args = render_args()
```
str4d, line 94:
```suggestion
    args = render_args()

    from random import Random
    rng = Random(0xabad533d)
    def randbytes(l):
        ret = []
        while len(ret) < l:
            ret.append(rng.randrange(0, 256))
        return bytes(ret)
    rand = Rand(randbytes)

```

orchard_key_components.py
str4d, line 105:
```suggestion
        sk = SpendingKey(rand.b(32))
        fvk = FullViewingKey(sk)
        note_v = rand.u64()
        note_r = Scalar.random(rand)
        note_rho = Fp.random(rand)
        note_psi = Fp.random(rand)
```

daira (commented):
utACK on the last three commits that I didn't co-author.

PR #32, opened 2021-05-28
Fix Orchard key components test vectors

str4d:
- Removed `note_r` (scalar copied from Sapling)
- Added `note_rseed` (bytes)
- `note_psi` needs to be derived from `note_rseed`
- missing `dk`, `note_rho` from output


therealyingtong (approved):

daira (approved):
utACK

PR #34, opened 2021-06-08
Add ZIP 244 auth digest test vectors

### LINKED ISSUE #33: Implement ZIP 244 auth digest test vectors
This was missed from #20.

str4d:
Closes #33.

### DIFF
--- a/zip_0244.py
+++ b/zip_0244.py
@@ -33,6 +33,12 @@ def transparent_digest(tx):
 
     return digest.digest()
 
+def transparent_scripts_digest(tx):
+    digest = blake2b(digest_size=32, person=b'ZTxAuthTransHash')
+    for x in tx.vin:
+        digest.update(bytes(x.scriptSig))
+    return digest.digest()
+
 # Sapling
 
 def sapling_digest(tx):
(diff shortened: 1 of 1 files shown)


zip_0244.py
```diff
@@ -45,6 +51,20 @@ def sapling_digest(tx):
 
     return digest.digest()
 
+def sapling_auth_digest(tx):
+    digest = blake2b(digest_size=32, person=b'ZTxAuthSapliHash')
+
+    if len(tx.vSpendsSapling) + len(tx.vOutputsSapling) > 0:
+        for desc in tx.vSpendsSapling:
+            digest.update(bytes(desc.proof))
```
nuttycom, line 59: Isn't `desc.proof` already a bytes value?

zip_0244.py
```diff
@@ -45,6 +51,20 @@ def sapling_digest(tx):
...
+        for desc in tx.vSpendsSapling:
+            digest.update(bytes(desc.spendAuthSig))
+        for desc in tx.vOutputsSapling:
+            digest.update(bytes(desc.proof))
```
nuttycom, line 63: Same question here.

nuttycom (approved):
Checked this against the spec. One question about some of the `bytes()` calls.

zip_0244.py
str4d: Not for Sapling (we have a `GrothProof` type).

zip_0244.py
str4d: See above.

ebfull (approved):

PR #35, opened 2021-06-10
Fix output for Orchard note encryption test vectors

str4d:
The test vectors were being generated correctly, but the output was still Sapling-specific (rather than including Orchard components).


orchard_note_encryption.py
```diff
@@ -252,17 +252,18 @@ def randbytes(l):
         assert(memo_using_ivk == memo)
 
         test_vectors.append({
+            'incoming_viewing_key': receiver_fvk.dk + bytes(ivk),
```
nuttycom, line 255: Will existing tests have to split this value into its parts? If so, would it be better to emit them as separate fields?

nuttycom (approved):
utACK modulo one uncertainty about how resulting data is used.

PR #37, opened 2021-06-10
Add Orchard empty root test vectors

str4d:


orchard_merkle_tree.py
```diff
@@ -11,6 +11,7 @@
 # https://zips.z.cash/protocol/nu5.pdf#constants
 MERKLE_DEPTH = 32
 L_MERKLE = 255
+UNCOMMITTED_ORCHARD = Fp(2)
```
therealyingtong, line 14: I thought we were using `Fp(0)` for uncommitted leaves: https://zcash.github.io/orchard/design/commitment-tree.html#uncommitted-leaves

str4d: The book is out of date here. We now use 0 as the mapping of the identity in ExtractP, so the uncommitted leaf is 2 (which is also not a valid x-coordinate).

str4d: https://zips.z.cash/protocol/protocol.pdf#thmuncommittedorchard

therealyingtong: Opened https://github.com/zcash/orchard/pull/112 to update the book description :+1:

PR #38, opened 2021-06-12
Add test vectors for unified address encodings.

### LINKED ISSUE #26: Implement Unified Address encodings

nuttycom:
This copies in the reference implementation of bech32m,
so that we can remove the 90 character length limitation.

Closes #26.

### DIFF
--- a/unified_addrs.py
+++ b/unified_addrs.py
@@ -127,10 +127,11 @@ def main():
         else: 
             orchard_raw_addr = None
 
+        is_p2pkh = rand.bool()
         receivers = [
             orchard_raw_addr, 
             sapling_raw_addr, 
-            (rand.bool(), t_addr)
+            (is_p2pkh, t_addr)
         ]
         ua = encode_unified(receivers)
 
@@ -140,7 +141,8 @@ def main():
         assert decoded.get('transparent') == t_addr
 
         test_vectors.append({
-            't_addr_bytes': t_addr,
+            'p2pkh_bytes': t_addr if is_p2pkh else None,
+            'p2sh_bytes': None if is_p2pkh else t_addr,
             'sapling_raw_addr': sapling_raw_addr,
             'orchard_raw_addr': orchard_raw_addr,
             'unified_addr': ua.encode()
@@ -150,7 +152,11 @@ def main():
         args,
         'unified_address',
         (
-            ('t_addr_bytes', {
+            ('p2pkh_bytes', {
+                'rust_type': 'Option<[u8; 20]>',
+                'rust_fmt': lambda x: None if x is None else Some(x),
+            }),
+            ('p2sh_bytes', {
                 'rust_type': 'Option<[u8; 20]>',
                 'rust_fmt': lambda x: None if x is None else Some(x),
             }),
@@ -170,4 +176,3 @@ def main():
 
 if __name__ == "__main__":
     main()
-
--- a/unified_addrs.py
+++ b/unified_addrs.py
@@ -1,0 +1,172 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+
+import math
+import struct
+
+from pyblake2 import blake2b
+from bech32m import bech32_encode, bech32_decode, convertbits, Encoding
+
+from tv_output import render_args, render_tv, Some
+from tv_rand import Rand
+from f4jumble import f4jumble, f4jumble_inv
+import sapling_key_components
+import orchard_key_components
+
+def encode_unified(receivers):
+    orchard_receiver = b""
+    if receivers['orchard']:
+        orchard_receiver = b"".join([b"\x03", len(receivers['orchard']).to_bytes(1, byteorder='little'), receivers['orchard']])
+
+    sapling_receiver = b""
+    if receivers['sapling']:
+        sapling_receiver = b"".join([b"\x02", len(receivers['sapling']).to_bytes(1, byteorder='little'), receivers['sapling']])
+
+    t_receiver = b""
+    if receivers['transparent']:
+        t_receiver = b"".join([b"\x01", len(receivers['transparent']).to_bytes(1, byteorder='little'), receivers['transparent']])
+
+    r_bytes = b"".join([orchard_receiver, sapling_receiver, t_receiver, bytes(16)])
+    converted = convertbits(f4jumble(r_bytes), 8, 5)
+    return bech32_encode("u", converted, Encoding.BECH32M)
+
+def decode_unified(addr_str):
+    (hrp, data, encoding) = bech32_decode(addr_str)
+    assert hrp == "u" and encoding == Encoding.BECH32M
+
+    decoded = f4jumble_inv(bytes(convertbits(data, 5, 8, False)))
+    suffix = decoded[-16:]
+    # check trailing zero bytes
+    assert suffix == bytes(16)
+    decoded = decoded[:-16]
+
+    s = 0
+    acc = []
+    result = {}
+    for b in decoded:
+        if s == 0:
+            receiver_type = b
+            assert [1, 2, 3].count(b) > 0, "receiver type " + str(b) + " not recognized"
+            s = 1
+        elif s == 1:
+            receiver_len = b
+            if receiver_type == 1:
+                assert receiver_len == 20
+            elif receiver_type == 2:
+                assert receiver_len == 43
+            elif receiver_type == 3:
+                assert receiver_len == 43
+            else:
+                assert False, "incorrect receiver length"
+            s = 2
+        elif s == 2:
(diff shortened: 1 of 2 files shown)


unified_addrs.py
```diff
@@ -0,0 +1,172 @@
...
+    s = 0
+    acc = []
+    result = {}
+    for b in decoded:
+        if s == 0:
+            receiver_type = b
+            assert [1, 2, 3].count(b) > 0, "receiver type " + str(b) + " not recognized"
+            s = 1
+        elif s == 1:
+            receiver_len = b
```
daira, line 52:
```
            expected_len == {1: 20, 2: 43, 3:43}.get(receiver_type)
            if expected_len is not None:
                assert receiver_len == expected_len, "incorrect receiver length"
```

nuttycom: Oh... right. :facepalm: 

Thanks

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+    if receivers['orchard']:
+        orchard_receiver = b"".join([b"\x03", len(receivers['orchard']).to_bytes(1, byteorder='little'), receivers['orchard']])
+
+    sapling_receiver = b""
+    if receivers['sapling']:
+        sapling_receiver = b"".join([b"\x02", len(receivers['sapling']).to_bytes(1, byteorder='little'), receivers['sapling']])
+
+    t_receiver = b""
+    if receivers['transparent']:
+        t_receiver = b"".join([b"\x01", len(receivers['transparent']).to_bytes(1, byteorder='little'), receivers['transparent']])
```
str4d, line 27: This is only exercising P2SH. We should also generate test vectors with typecode `0x00` for P2PKH (which is likely to be more common). Since they are exclusive, and their values are the same length, we can just randomly pick one typecode or the other.

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+        if s == 0:
+            receiver_type = b
+            assert [1, 2, 3].count(b) > 0, "receiver type " + str(b) + " not recognized"
+            s = 1
+        elif s == 1:
+          receiver_len = b
+          expected_len == {1: 20, 2: 43, 3: 43}.get(receiver_type)
+            if expected_len is not None:
+                assert receiver_len == expected_len, "incorrect receiver length"
+        elif s == 2:
```
str4d, line 56: How is this case ever triggered? `s` is only ever set to either `0` or `1`.

EDIT: Ah, the `s = 2` state transition was accidentally removed in the second commit. Add it back in:
```suggestion
            s = 2
        elif s == 2:
```

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+    assert suffix == bytes(16)
+    decoded = decoded[:-16]
+
+    s = 0
+    acc = []
+    result = {}
+    for b in decoded:
+        if s == 0:
+            receiver_type = b
+            assert [1, 2, 3].count(b) > 0, "receiver type " + str(b) + " not recognized"
```
str4d, line 49: This is rejecting unknown receiver types.
```suggestion
```

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+            s = 1
+        elif s == 1:
+          receiver_len = b
+          expected_len == {1: 20, 2: 43, 3: 43}.get(receiver_type)
```
str4d, line 53: This is missing the P2PKH typecode:
```suggestion
          expected_len == {0: 20, 1: 20, 2: 43, 3: 43}.get(receiver_type)
```

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+                assert receiver_len == expected_len, "incorrect receiver length"
+        elif s == 2:
+            if len(acc) < receiver_len:
+                acc.append(b)
+            
+            if len(acc) == receiver_len:
+                if receiver_type == 1:
+                    assert not ('transparent' in result), "duplicate transparent receiver detected"
+                    assert len(acc) == 20
+                    result['transparent'] = bytes(acc)
```
str4d, line 64: This needs to handle `receiver_type == 0`, and _also_ ensure that we only see either `0x00` or `0x01`.

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+            orchard_raw_addr = b"".join([orchard_default_d[:11], bytes(orchard_default_pk_d)[:32]])
+        else: 
+            orchard_raw_addr = None
+
+        receivers = {
+            'orchard': orchard_raw_addr, 
+            'sapling': sapling_raw_addr, 
+            'transparent': t_addr
+        }
+        ua = encode_unified(receivers)
```
str4d, line 128: Using a dictionary for this makes the precise ordering of the UA internals non-deterministic (because Python uses a hashmap for dictionaries), which means non-deterministic test vectors. Use a list of tuples instead:
```suggestion
        receivers = [
            ('orchard', orchard_raw_addr),
            ('sapling', sapling_raw_addr),
            ('transparent', t_addr),
        ]
        ua = encode_unified(receivers)
```

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+            'orchard_raw_addr': orchard_raw_addr,
+            'unified_addr': ua.encode()
+        })
+
+    render_tv(
+        args,
+        'unified_address',
+        (
+            ('t_addr_bytes', {
+                'rust_type': 'Option<[u8; 22]>',
```
str4d, line 147:
```suggestion
                'rust_type': 'Option<[u8; 20]>',
```

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+                'rust_fmt': lambda x: None if x is None else Some(x),
+            }),
+            ('sapling_raw_addr', {
+                'rust_type': 'Option<[u8; 22]>',
```
str4d, line 151:
```suggestion
                'rust_type': 'Option<[u8; 43]>',
```

unified_addrs.py
```diff
@@ -0,0 +1,166 @@
...
+                'rust_fmt': lambda x: None if x is None else Some(x),
+            }),
+            ('orchard_raw_addr', {
+                'rust_type': 'Option<[u8; 22]>',
```
str4d, line 155:
```suggestion
                'rust_type': 'Option<[u8; 43]>',
```

str4d (changes requested):

unified_addrs.py
nuttycom: The duplicate transparent receiver check will suffice, given the change to `receiver_type == 0 or receiver_type == 1`

unified_addrs.py
```diff
@@ -0,0 +1,173 @@
...
+        ]
+        ua = encode_unified(receivers)
+
+        decoded = decode_unified(ua)
+        assert decoded.get('orchard') == orchard_raw_addr
+        assert decoded.get('sapling') == sapling_raw_addr
+        assert decoded.get('transparent') == t_addr
+
+        test_vectors.append({
+            't_addr_bytes': t_addr,
```
str4d, line 143: To use these test vectors, we need to know what kind of t-addr is expected:
```suggestion
            'p2pkh_bytes': t_addr if is_p2pkh else None,
            'p2sh_bytes': None if is_p2pkh else t_addr,
```

unified_addrs.py
```diff
@@ -0,0 +1,173 @@
...
+            orchard_default_pk_d = orchard_fvk.default_pkd()
+            orchard_raw_addr = b"".join([orchard_default_d[:11], bytes(orchard_default_pk_d)[:32]])
+        else: 
+            orchard_raw_addr = None
+
+        receivers = [
+            orchard_raw_addr, 
+            sapling_raw_addr, 
+            (rand.bool(), t_addr)
+        ]
```
str4d, line 134:
```suggestion
        is_p2pkh = rand.bool()
        receivers = [
            orchard_raw_addr, 
            sapling_raw_addr, 
            (is_p2pkh, t_addr)
        ]
```

unified_addrs.py
```diff
@@ -0,0 +1,173 @@
...
+        })
+
+    render_tv(
+        args,
+        'unified_address',
+        (
+            ('t_addr_bytes', {
+                'rust_type': 'Option<[u8; 20]>',
+                'rust_fmt': lambda x: None if x is None else Some(x),
+            }),
```
str4d, line 156:
```suggestion
            ('p2pkh_bytes', {
                'rust_type': 'Option<[u8; 20]>',
                'rust_fmt': lambda x: None if x is None else Some(x),
            }),
            ('p2sh_bytes', {
                'rust_type': 'Option<[u8; 20]>',
                'rust_fmt': lambda x: None if x is None else Some(x),
            }),
```

str4d (approved):
ACK after these changes.

PR #39, opened 2021-06-23
Add vector generation for bundle commitments.

nuttycom:
This is a bit awkward, because the test vectors it generates are not
complete; it merely generates anchors for the bundle commitments
from https://github.com/zcash/zcash/pull/5233/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&file-filters%5B%5D=.hpp#diff-38ff85a72cad7fe2516912a84c7deedb725cc444d230a6d1accfc1258ecff005

The serialized bundle values there were generated in a non-reproducible
fashion from the librustzcash property tests, then the note commitments
were extracted so as to be able to generate just the anchors.


therealyingtong:
This is currently conflicting with `master`.

orchard_merkle_tree.py
```diff
@@ -32,9 +73,253 @@ def merkle_crh(layer, left, right):
...
-def empty_roots():
-    empty_roots = [UNCOMMITTED_ORCHARD]
-    for layer in range(0, MERKLE_DEPTH)[::-1]:
-        bits = i2lebsp(L_MERKLE, empty_roots[-1].s)
-        empty_roots.append(merkle_crh(layer, bits, bits))
-    return empty_roots
+def main():
+    args = render_args()
+
+    bundle_commitments = [
```
str4d, line 79: This should not be in this file (this is where #36 will go). These should go into a separate script file (they aren't really what this repo is for, but there isn't any better place to put them).

str4d: Document how these were generated, in case they need to be regenerated.

nuttycom:
Once #74 is complete, this should be closed without merging.

PR #40, opened 2021-06-25
Add a requirements.txt pip file compatible with Python 3

dconnolly:


daira (approved):
utACK

requirements.txt
```diff
@@ -0,0 +1,6 @@
+cffi==1.14.5
```
str4d, line 1: This is not a direct dependency, but I confirmed that it is pulled in by `cryptography`, so I'm fine with it being pinned here.

requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+chacha20poly1305==0.0.3
+cryptography==3.4.7
+numpy==1.21.0
+pyblake2==1.1.2
+pycparser==2.20
```
str4d, line 6: This is not a direct dependency, but I confirmed that it is pulled in by `cryptography`, so I'm fine with it being pinned here.

str4d (approved):
Tested ACK.

PR #45, opened 2021-08-04
Update Poseidon instantiation from 58 to 56 partial rounds.

daira:
refs https://github.com/zcash/orchard/issues/166


defuse:
utACK I re-ran the sage script and verified all the constants.

str4d (approved):
utACK. I did not check the constants; I am relying on these test vectors passing against the `orchard` impl, and the `orchard` impl's independent implementation of `generate_parameters_grain.sage` in RUst.

PR #49, opened 2021-08-30
Updates for large UAs/UVKs

daira:
This supports the spec changes in https://github.com/zcash/zips/pull/551 .


zc_utils.py
```diff
@@ -51,4 +60,7 @@ def test_round_trip(n, encoding):
 test_round_trip(0xFFFFFFFE, b'\xFE\xFE\xFF\xFF\xFF')
 test_round_trip(0xFFFFFFFF, b'\xFE\xFF\xFF\xFF\xFF')
 test_round_trip(0x0100000000, b'\xFF\x00\x00\x00\x00\x01\x00\x00\x00')
-test_round_trip(0xFFFFFFFFFFFFFFFF, b'\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF\xFF')
```
str4d: Why was this case removed?

f4jumble.py
```diff
@@ -123,7 +136,17 @@ def randbytes(l):
...
+    print()
+    render_tv(
+        args,
+        'f4jumble',
+        (
+            ('length', 'usize'),
+            ('jumbled_hash', 'Vec<u8>'),
+        ),
+        hashed_test_vectors,
     )
```
str4d, line 150: There is only supposed to be one set of test vectors returned per file. Move this to a new test file `f4jumble_hashed.py`.

str4d (changes requested):

daira:
BTW when the GitHub UI says "Empty file." for a file with changed permissions, that's wrong; it just means that the file contents have not changed.

zc_utils.py
daira: I don't remember. I'll restore it.

zc_utils.py
```diff
@@ -1,3 +1,6 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+
```
str4d, line 3: Weakly: this file does not generate test vectors, so it shouldn't be executable. (The assertions are run when this file is imported by others.)

zc_utils.py
```diff
@@ -10,18 +13,58 @@ def write_compact_size(n):
...
+    elif b == 254:
+        assert len(rest) >= 5
+        n = struct.unpack('<I', rest[1:5])[0]
+        assert n >= 0x10000
+        return (n, rest[5:])
+    else:
+        assert len(rest) >= 9
+        n = struct.unpack('<Q', rest[1:9])[0]
+        assert n >= 0x100000000
+        return (n, rest[9:])
```
str4d, line 35: The parser should also (perhaps optionally) enforce the max size limit (so at the parse callsite we can enforce the intended behaviour).

f4jumble_long.py
```diff
@@ -0,0 +1,41 @@
...
+            'length': l_M,
+            'jumbled_hash': blake2b(jumbled).digest()
+        })
+
+    render_tv(
+        args,
+        'f4jumble_long',
+        (
+            ('length', 'usize'),
+            ('jumbled_hash', 'Vec<u8>'),
```
str4d, line 34:
```suggestion
            ('jumbled_hash', '[u8; 64]'),
```

unified_address.py
```diff
@@ -51,46 +52,35 @@ def decode_unified(addr_str):
...
-
-                elif receiver_type == 3:
-                    assert not ('orchard' in result), "duplicate orchard receiver detected"
-                    assert len(acc) == 43
-                    result['orchard'] = bytes(acc)
-                    acc = []
-                    s = 0
+    while len(rest) > 0:
+        (receiver_type, rest) = parse_compact_size(rest)
+        (receiver_len, rest) = parse_compact_size(rest)
```
str4d, line 60: We should be enforcing the maximum CompactSize length here (by passing the appropriate flag, or having `parse_compact_size` default to enforcing it).

unified_address.py
```diff
@@ -4,17 +4,18 @@
...
 from tv_rand import Rand
+from zc_utils import write_compact_size, parse_compact_size
 from f4jumble import f4jumble, f4jumble_inv
 import sapling_key_components
 import orchard_key_components
 
+
 def tlv(typecode, value):
-    return b"".join([bytes([typecode, len(value)]), value])
+    return b"".join([write_compact_size(typecode), write_compact_size(len(value)), value])
```
str4d, line 18: Similarly, we should be asserting inside `write_compact_size` that these are within the correct CompactSize range (either explicitly, or by default).

str4d (changes requested):
The CompactSize parser change is blocking.

nuttycom (approved):
ACK, these are now in use so this should be merged.

zc_utils.py
daira: I disagree; it's useful to be able to run the assertions in this file separately.

str4d (approved):

PR #50, opened 2021-08-31
orchard_vesta.py: Add constants for Vesta curve.

therealyingtong:
I wanted to extract general `Point` and `Field` classes across Pallas/Vesta, but the changes got too involved.


daira (approved):
utACK. Verified all constants and that all p↔q changes have been done correctly.

PR #51, opened 2021-08-31
Add Fq test vectors for Poseidon.

therealyingtong:
Based on #50.


orchard_poseidon/hash/fp.py
```diff
@@ -1,5 +1,9 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+sys.path.append('../..')
```
daira, line 3:
```suggestion
sys.path.insert(1, os.path.join(sys.path[0], '../..'))
```
This is the recommended pattern because it works:
* even if the script was not invoked from the current directory;
* even if there are modules on `sys.path` that could accidentally include items with the same names.

(Found [here](https://stackoverflow.com/questions/714063/importing-modules-from-parent-folder#comment23054549_11158224).)

orchard_poseidon/hash/fq.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+sys.path.append('../..')
```
daira, line 3:
```suggestion
import os; sys.path.insert(1, os.path.join(sys.path[0], '../..'))
```

orchard_poseidon/permute/fp.py
```diff
@@ -1,3 +1,7 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+sys.path.append('../..')
```
daira, line 3:
```suggestion
import os; sys.path.insert(1, os.path.join(sys.path[0], '../..'))
```

orchard_poseidon/permute/fq.py
```diff
@@ -0,0 +1,201 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+sys.path.append('../..')
```
daira, line 3:
```suggestion
import os; sys.path.insert(1, os.path.join(sys.path[0], '../..'))
```

daira (approved):
utACK

PR #53, opened 2021-09-17
Ensure that GitHub's "View license" link points to complete license information

daira:
refs https://github.com/zcash/librustzcash/issues/443


steven-ecc:
I don't see the "View License" link on the right. Should that be present for a branch as well?

image

daira:
No, that depends on what is in the default branch (as it should, because it would be misleading to show a different license for a PR that changes license files in a way that hasn't been approved by the repo maintainers).

steven-ecc (approved):
utACK

therealyingtong (approved):
ACK.

README.md
```diff
@@ -9,16 +9,4 @@ Requirements:
...
-
-at your option.
-
-### Contribution
-
-Unless you explicitly state otherwise, any contribution intentionally
-submitted for inclusion in the work by you, as defined in the Apache-2.0
-license, shall be dual licensed as above, without any additional terms or
-conditions.
+This repository is [MIT/Apache 2.0 dual-licensed](COPYING).
```
str4d, line 12: The existing preamble above is an ecosystem-wide standard for Rust crates, because what crates.io shows is the `README`, and the `Cargo.toml`'s `license` field, not any `COPYING` file. I would greatly prefer that the above preamble is left exactly as-is, and we instead duplicate it in `COPYING` for GitHub's benefit.

str4d: (I say this here because @daira suggested elsewhere duplicating this pattern; I'm ambivalent about this pattern here in Python code.)

daira: I will revert this and update the other PRs.

daira: Addressed in #60.

PR #54, opened 2021-09-28
Updates for new NU5 testnet consensus rules

str4d:
The ZIP 244 test vectors are altered as a result of changing the
consensus branch ID.


nuttycom (approved):
utACK

daira (commented):
Post-hoc utACK

PR #61, opened 2022-01-06
Update test vectors for ZIP 244 changes

### LINKED ISSUE #57: ZIP 244: Implement changes to bring its transparent semantics closer to BIP 341
See https://github.com/zcash/zips/issues/574#issuecomment-1002803827

str4d:
Closes #57.

### DIFF
--- a/zip_0244.py
+++ b/zip_0244.py
@@ -221,35 +221,34 @@ def auth_digest(tx):
 # Signatures
 
 class TransparentInput(object):
-    def __init__(self, tx, rand):
+    def __init__(self, nIn, rand):
+        self.nIn = nIn
         self.scriptCode = Script(rand)
-        self.nIn = rand.u8() % len(tx.vin)
         self.amount = rand.u64() % (MAX_MONEY + 1)
 
-def signature_digest(tx, nHashType, txin):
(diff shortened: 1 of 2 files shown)


daira (approved):
utACK

zip_0244.py
```diff
@@ -357,19 +400,20 @@ def randbytes(l):
             ('tx', {'rust_type': 'Vec<u8>', 'bitcoin_flavoured': False}),
             ('txid', '[u8; 32]'),
             ('auth_digest', '[u8; 32]'),
+            ('amounts', '[i64; %d]' % len(t_inputs)),
+            ('script_codes', {
+                'rust_type': '[Vec<u8>; %d]' % len(t_inputs),
```
str4d, line 405: Kris noted that this causes a bug in the output, because `len(t_inputs)` returns the length for the _last_ test vector, but it varies across them. This should be:
```suggestion
            ('amounts', 'Vec<i64>'),
            ('script_codes', {
                'rust_type': 'Vec<Vec<u8>>',
```
but that won't work without more changes to the Rust output renderer, so instead I'm going to overhaul the latter after this PR.

PR #62, opened 2022-01-07
Refactor test vectors into package

str4d:
Scripts that were previously run via:
```
poetry run python ./SCRIPT_NAME.py
```

and have been moved, can now be run as:
```
poetry run SCRIPT_NAME
```

Some top-level scripts remain, where they only generate test vectors and do not contain any Zcash logic. These now use the `zcash_test_vectors` package, and can't be run directly unless this package is made available in `PYTHONPATH`.


nuttycom (changes requested):
This looks great overall. With these changes, however, the README is out of date - it would be nice to update it to clarify how to run test vector generation (in particular, those at the top level that you mention require making the package available in PYTHONPATH, since I'm not sure what exactly this means).

str4d:
> With these changes, however, the README is out of date

I thought it made no mention of how to run things at all, but I see it does refer to two specific test vectors by name. I'll fix those, and add a general description of how to run things.

> I'm not sure what exactly this means

What I mean there is, we can no longer run these directly outside of `poetry`. `poetry` installs the package in its venv, so things Just Work, but someone trying to run those scripts directly will have issues.

What I'm planning to do in a subsequent PR is refactor the test vector generators, so we _can_ have multiple TV generators in each file (whereas previously we needed one file per TV set).

daira (approved):
utACK

nuttycom (approved):
utACK

daira (commented):
Post-hoc ACK

PR #63, opened 2022-01-07
CI: Add workflow to verify test vectors are up-to-date

str4d:
Also commits all test vectors we can generate.


nuttycom:
I think that something's missing from this commit; I attempted to reproduce the test vectors and get this: ```
nuttycom@sword:~/work/zcash-test-vectors (ci-check-test-vectors)λ ./regenerate.sh
# f4jumble_long
usage: f4jumble_long.py [-h] [-t {zcash,json,rust}]
f4jumble_long.py: error: argument -t/--target: expected one argument
# f4jumble
usage: f4jumble.py [-h] [-t {zcash,json,rust}]
f4jumble.py: error: argument -t/--target: expected one argument
# orchard_empty_roots
usage: orchard_empty_roots.py [-h] [-t {zcash,json,rust}]
orchard_empty_roots.py: error: argument -t/--target: expected one argument
# orchard_generators
usage: orchard_generators.py [-h] [-t {zcash,json,rust}]
orchard_generators.py: error: argument -t/--target: expected one argument
# orchard_group_hash
usage: orchard_group_hash.py [-h] [-t {zcash,json,rust}]
orchard_group_hash.py: error: argument -t/--target: expected one argument
...
```

str4d:
See the workflow; you need to supply arguments: `./regenerate.sh rust rs` etc. At one point I had the script internalise them and do all three, but it was going to make CI three times slower, so now the CI passes those args.

str4d:
The error in the JSON workflow was expected by me because I'd seen it locally (turns out there's a bug in the `orchard_merkle_tree.py` test vector generator that means it can generate Rust but not JSON), but the CI should continue to check the other two. Fixing...

str4d:
Excellent, CI is detecting the bad JSON serializations, but confirming the Rust test vectors all run fine and match the checked-in data. Opened #64 for fixing the JSON output for that test.

nuttycom (approved):
ACK. I have verified that the test vectors being committed here are the same as those I was able to generate locally.

therealyingtong:
Do we expect the `Verify there are no changes` workflow to fail for randomly generated test vectors?

daira:
@therealyingtong All "randomly" generated test vectors use a fixed RNG seed, so they are deterministic.

str4d:
Yep. The use of a seeded RNG enables people to generate many more test vectors if they want; we just hard-code a default seed and test vector count that results in reasonable space coverage.

PR #68, opened 2022-01-13
Add generated test vectors for zip_143, zip_243, zip_244, unified add…

nuttycom:
…resses, and ufvks


regenerate.sh
```diff
@@ -17,7 +17,13 @@ tv_scripts=(
...
     sapling_note_encryption
     sapling_signatures
-    sapling_zip32)
+    sapling_zip32
+    unified_address
+    unified_full_viewing_keys
+    zip_0143
+    zip_0243
+    zip_0244
+    )
```
str4d, line 26: This will break the script (in a way that isn't detected until the script ends), because it inserts an empty-string `""` element at the end of the array. Remove the extra newline:
```suggestion
    zip_0244)
```

str4d (changes requested):

str4d (approved):
ACK. I verified that I generated the same Rust and JSON test vectors (I didn't check the Bitcoin-flavoured ones).

daira (commented):
Post-hoc ACK

PR #69, opened 2022-01-13
`unified_full_viewing_keys.py`: Provide default_backend() to ec.derive_private_key().

therealyingtong:


therealyingtong:
Ah, this isn't a problem if I run `poetry run python3 ./unified_full_viewing_keys.py`.

PR #70, opened 2022-01-13
Add test vectors for unified IVKs

### LINKED ISSUE #56: Unified IVK test vectors.
Similar to #47

therealyingtong:
Closes #56.


daira (approved):
utACK. I made a few minor simplifications and formatting changes, and I removed the .json files from this PR (they will be added in #77 which fixes the orchard_merkle_tree test vector serialization). The JSON and Bitcoin-flavoured CI checks are expected to fail in this PR.

unified_incoming_viewing_keys.py
```diff
@@ -0,0 +1,126 @@
...
+            t_key_bytes = c + pubkey_bytes
+        else:
+            t_key_bytes = None
+
+        has_s_key = rand.bool()
+        if has_s_key:
+            root_key = sapling_zip32.ExtendedSpendingKey.master(seed)
+            purpose_key = root_key.child(sapling_zip32.hardened(32))
+            coin_key = purpose_key.child(sapling_zip32.hardened(133))
+            account_key = coin_key.child(sapling_zip32.hardened(i))
```
nuttycom, line 42: Post-hoc: why does the Sapling key use correct ZIP-316 derivation, while the Orchard and transparent components just use random keys?

nuttycom (commented):
Post-hoc question.

unified_incoming_viewing_keys.py
daira: I think it's because there is no implementation in this repo of master key generation for Orchard or transparent. The same comment applies to `unified_full_viewing_keys.py`. I'll consider fixing this when I implement the Orchard ZIP 32 test vectors.

PR #72, opened 2022-01-25
Update ZIP 244 test vectors to reflect changes in coinbase transaction signatures.

nuttycom:


test-vectors/rust/zip_0244.rs
```diff
@@ -55,21 +55,21 @@
...
..., 0x31, 0xa1, 0xbd, 0x62, 0x81, 0xae, 0xdf, 0x2a, 0x0e, 0x8f, 0xab, 0x18, 0x81, 0x72, 0x02, 0xa9, 0xbe, 0x06, 0x40, 0x2e, 0xd9, 0xcc, 0x72, 0x0c, 0x16, 0xbf, 0xe8, 0x81, 0xe4, 0xdf, 0x42, 0x55, 0xe8, 0x7a, 0xfb, 0x7f, 0xc6, 0x2f, 0x38, 0x11, 0x6b, 0xbe, 0x03, 0xcd, 0x8a, 0x3c, 0xb1, 0x1a, 0x27, 0xd5, 0x68, 0x41, 0x47, 0x82, 0xf4, 0x7b, 0x1a, 0x44, 0xc9, 0x7c, 0x68, 0x04, 0x67, 0x69, 0x4b, 0xc9, 0x70, 0x9d, 0x32, 0x91, 0x6c, 0x97, 0xe8, 0x00, 0x6c, 0xbb, 0x07, 0xba, 0x0e, 0x41, 0x80, 0xa3, 0x73, 0x80, 0x38, 0xc3, 0x74, 0xc4, 0xcc, 0xe8, 0xf3, 0x29, 0x59, 0xaf, 0xb2, 0x5f, 0x30, 0x3f, 0x58, 0x15, 0xc4, 0x53, 0x31, 0x24, 0xac, 0xf9, 0xd1, 0x89, 0x40, 0xe7, 0x75, 0x22, 0xac, 0x5d, 0xc4, 0xb9, 0x57, 0x0a, 0xae, 0x8f, 0x47, 0xb7, 0xf5, 0x7f, 0xd8, 0x76, 0x7b, 0xea, 0x1a, 0x24, 0xae, 0x7b, 0xed, 0x65, 0xb4, 0x09, 0xe1, 0xdd, 0x26, 0xb8, 0xdd, 0xdd, 0x68, 0x85, 0x8d, 0x6f, 0x51, 0x61, 0xf0, 0x73, 0xd9, 0x06, 0x36, 0x86, 0x0a, 0x9a, 0xae, 0xe1, 0x86, 0x29, 0xb0, 0x63, 0x30, 0xa8, 0xee, 0x30, 0x59, 0x1d, 0xeb, 0xfc, 0xef, 0x56, 0xa0, 0x26, 0xbb, 0x28, 0xc3, 0xb0, 0x6e, 0xc2, 0xcf, 0xaf, 0x5b, 0x79, 0xab, 0x72, 0x69, 0x4d, 0x1d, 0x01, 0x2a, 0x75, 0x94, 0xdd, 0x80, 0xae, 0x7d, 0xfa, 0x0c, 0x00
```
daira, line 58: There are now three coinbase transactions (out of 10); I checked that there are still some remaining non-coinbase transactions with zero inputs.

daira:
The CI checks for the JSON and Bitcoin-flavoured vectors are expected to fail.

zcash_test_vectors/transaction.py
```diff
@@ -385,6 +412,11 @@ def __init__(self, rand, consensus_branch_id):
         if have_transparent_in:
             for _ in range((rand.u8() % 3) + 1):
                 self.vin.append(TxIn(rand))
+        if is_coinbase:
+            self.vin.append(TxIn.from_components(
+                OutPoint.from_components(b'\x00' * 32, 0xFFFFFFFF),
+                Script.from_bytes(b"\x51"),
```
str4d, line 418: To match real coinbase transactions, this should be a valid height encoding (which the current opcode is), followed by `OP_0`:
```suggestion
                Script.from_bytes(b"\x51\x00"), // [INT64(1), OP_0]
```

zcash_test_vectors/transaction.py
```diff
@@ -385,6 +412,11 @@ def __init__(self, rand, consensus_branch_id):
...
+                0))
```
str4d, line 419: The default `nSequence` is `std::numeric_limits<unsigned int>::max()`, and `src/miner.cpp` doesn't alter it.
```suggestion
                0xffffffff))
```

str4d (changes requested):
utACK after these changes to fixed values within coinbase transactions.

zcash_test_vectors/transaction.py
daira: The intention was to encode a height of 1, since there was no height in the genesis coinbase transaction. [According to the spec](https://zips.z.cash/protocol/protocol.pdf#txnconsensus) this should be [0x01, 0x51] (I had forgotten the length).

image

str4d (approved):

PR #75, opened 2022-01-31
Add test vectors for internal key derivation

daira:


test-vectors/rust/orchard_key_components.rs
```diff
@@ -9,6 +9,12 @@
             dk: [u8; 32],
             default_d: [u8; 11],
             default_pk_d: [u8; 32],
+            internal_ak: [u8; 32],
+            internal_nk: [u8; 32],
```
str4d, line 13: These are identical to `ak` and `nk`, so we should just use those values.

zcash_test_vectors/orchard/key_components.py
```diff
@@ -54,7 +54,11 @@ def __init__(self, data):
 
 class FullViewingKey(object):
     def __init__(self, sk):
-        (self.rivk, self.ak, self.nk) = (sk.rivk, sk.ak, sk.nk)
+        if isinstance(sk, SpendingKey):
+            (self.rivk, self.ak, self.nk) = (sk.rivk, sk.ak, sk.nk)
+        else:
+            (self.rivk, self.ak, self.nk) = sk
```
str4d, line 60: This would be clearer if we made the constructor take the three components (`(self, ak, nk, rivk)`), and then had two class methods for the different construction cases.

zcash_test_vectors/orchard/key_components.py
```diff
@@ -120,6 +130,12 @@ def randbytes(l):
             'dk': fvk.dk,
             'default_d': default_d,
             'default_pk_d': bytes(default_pk_d),
+            'internal_ak': bytes(internal.ak),
+            'internal_nk': bytes(internal.nk),
```
str4d, line 134:
```suggestion
```

zcash_test_vectors/orchard/key_components.py
```diff
@@ -141,6 +157,12 @@ def randbytes(l):
             ('dk', '[u8; 32]'),
             ('default_d', '[u8; 11]'),
             ('default_pk_d', '[u8; 32]'),
+            ('internal_ak', '[u8; 32]'),
+            ('internal_nk', '[u8; 32]'),
```
str4d, line 161:
```suggestion
```

str4d (approved):
utACK with simplification suggestions.

nuttycom (approved):
utACK. I agree with @str4d's nitpick about the Orchard FullViewingKey constructor.

therealyingtong (approved):
utACK

zcash_test_vectors/sapling/zip32.py
```diff
@@ -111,6 +121,19 @@ def child(self, i):
...
 
+    def internal(self):
+        FVK   = encode_xfvk_parts(self.ak(), self.nk(), self.ovk(), self.dk())
+        I     = blake2b(person=b'Zcash_SaplingInt', digest_size=32, data=FVK).digest()
+        I_nsk = to_scalar(prf_expand(I, b'\x17'))
+        R     = prf_expand(I, b'\x18')
+        nsk_internal = I_nsk + self.nsk()
+        dk_internal  = R[:32]
+        ovk_internal = R[32:]
+        internal = self.__class__(self.ask(), nsk_internal, ovk_internal, dk_internal, self._c, self.depth(), self.tag(), self._i)
```
therealyingtong, line 132: Should this use `self.parent_tag()` since the class constructor expects a `parent_tag` argument?
```suggestion
        internal = self.__class__(self.ask(), nsk_internal, ovk_internal, dk_internal, self._c, self.depth(), self.parent_tag(), self._i)
```

therealyingtong (dismissed):

zcash_test_vectors/sapling/zip32.py
```diff
@@ -170,10 +203,22 @@ def child(self, i):
...
-def hardened(i): 
+    def internal(self):
+        FVK   = encode_xfvk_parts(self.ak(), self.nk(), self.ovk(), self.dk())
+        I     = blake2b(person=b'Zcash_SaplingInt', digest_size=32, data=FVK).digest()
+        I_nsk = to_scalar(prf_expand(I, b'\x17'))
+        R     = prf_expand(I, b'\x18')
+        nk_internal  = PROVING_KEY_BASE * I_nsk + self.nk()
+        dk_internal  = R[:32]
+        ovk_internal = R[32:]
+        return self.__class__(self.ak(), nk_internal, ovk_internal, dk_internal, self._c, self.depth(), self.tag(), self._i)
```
therealyingtong, line 214: Similarly here.

```suggestion
        return self.__class__(self.ak(), nk_internal, ovk_internal, dk_internal, self._c, self.depth(), self.parent_tag(), self._i)
```

test-vectors/rust/sapling_zip32.rs
```diff
@@ -14,6 +14,12 @@
             d1: Option<[u8; 11]>,
             d2: Option<[u8; 11]>,
             dmax: Option<[u8; 11]>,
+            internal_nsk: Option<[u8; 32]>,
+            internal_ovk: [u8; 32],
+            internal_dk: [u8; 32],
+            internal_nk: [u8; 32],
+            internal_xsk: Option<[u8; 169]>,
+            internal_xfvk: [u8; 169],
```
therealyingtong, line 22: Do we also want to include `internal_fp`, `internal_ivk`?

zcash_test_vectors/orchard/key_components.py
daira: We only need a single class method, `FullViewingKey.from_spending_key`, because the other case can use the constructor directly. Done.

nuttycom (approved):
re-utACK

therealyingtong (approved):
utACK

PR #80, opened 2022-02-08
Move remaining scripts out of the root directory

daira:
closes #79

Signed-off-by: Daira Hopwood <daira@jacaranda.org>

PR #81, opened 2022-02-08
ZIP 32 test vectors for Orchard and BIP 32 test vectors for transparent

### LINKED ISSUE #41: Orchard ZIP32 test vectors

daira:
closes #41


daira:
This should be merged after #80, when it is finished.

daira:
I suggest reviewing this commit-by-commit.

nuttycom (approved):
utACK

zcash_test_vectors/unified_address.py
```diff
@@ -21,21 +22,29 @@ def main():
...
             t_addr = None
 
+        j = 0
         has_s_addr = rand.bool()
         if has_s_addr:
-            sapling_sk = sapling_key_components.SpendingKey(rand.b(32))
-            sapling_default_d = sapling_sk.default_d()
-            sapling_default_pk_d = sapling_sk.default_pkd()
-            sapling_raw_addr = b"".join([sapling_default_d[:11], bytes(sapling_default_pk_d)[:32]])
+            rand.b(32) # discard
```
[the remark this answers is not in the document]
daira: The commits are written in the following order:
1. trivial refactor
2. changes only the Sapling vectors
3. changes only the Orchard vectors
4. changes only the transparent vectors
5. cleanup that ensures we are using the randomness in precisely the same way for unified_address, unified_incoming_viewing_keys, and unified_full_viewing_keys (so corresponding test vectors are about the same keys)
6. optimization

In commits 2, 3, and 4, we're careful to retain all of the sampling of randomness, even if it is now discarded (because otherwise all of the test vectors after that point would change). This is because @str4d wanted to be able to verify commit-by-commit that only the expected vectors had changed. However, then I realised that making the test vectors "match up" between unified_incoming_viewing_keys, and unified_full_viewing_keys would change all the vectors at once in commit 5, anyway. But I still think this approach made it easier for me to check that the vectors were only changing in expected ways (even if it took a long time to get it right).

zcash_test_vectors/unified_full_viewing_keys.py
```diff
@@ -100,6 +111,7 @@ def main():
             'unknown_fvk_typecode': unknown_tc,
```
nuttycom, line 111: Please also include the seed, so that this can be used to test derivation the whole way through.

nuttycom: Added https://github.com/daira/zcash-test-vectors/pull/1 for this.

daira: The seed is fixed (00 01 .. 1f) and doesn't change between test vector entries; do you want me to add it to all entries?

zcash_test_vectors/unified_full_viewing_keys.py
```diff
@@ -3,45 +3,54 @@
...
-            assert pubkey_bytes[0] in (0x02, 0x03)
-            t_key_bytes = c + pubkey_bytes
+            rand.b(20) # discard, to match UA generation
+
+            # <https://zips.z.cash/zip-0316#encoding-of-unified-full-incoming-viewing-keys>
+            # "However, the [Transparent P2PKH] FVK uses the key at the Account level, i.e.
+            # at path m/44'/coin_type'/account', while the IVK uses the external (non-change)
+            # child key at the Change level, i.e. at path m/44'/coin_type'/account'/0."
+            t_account_key = t_coin_key.child(hardened(account))
+            t_key_bytes = bytes(t_account_key.public_key())
```
nuttycom, line 46: Blocking: I believe that this is missing the chain code.

nuttycom: Edit: okay, this is not the public key, it's the extended public key, which has the correct byte representation.

daira:
After this PR, the zip_0316 test vectors are arguably redundant with the bip_0044 ones. Do yous want me to remove the former and simplify by moving [derive_ovks](https://github.com/zcash-hackworks/zcash-test-vectors/blob/cc3376f37ecf9cddbaabd593affa3fda5db40873/zcash_test_vectors/transparent/zip_0316.py#L10-L15) into `bip_0044.py`?

zcash_test_vectors/unified_address.py
```diff
@@ -139,6 +140,7 @@ def main():
                 'rust_fmt': lambda x: None if x is None else Some(x),
             }),
             ('unified_addr', 'Vec<u8>'),
+            ('root_seed', 'Vec<u8>'),
```
daira, line 143: ('unified_addr', {'rust_type': 'Vec<u8>', 'bitcoin_flavoured': False}),
            ('root_seed', {'rust_type': 'Vec<u8>', 'bitcoin_flavoured': False}),

zcash_test_vectors/transparent/bip_0032.py
```diff
@@ -0,0 +1,123 @@
...
+    def public_key(self):
+        return ExtendedPublicKey(self.sk.pubkey, self.chaincode)
+
+    def child(self, i):
+        assert 0 <= i and i <= 0xFFFFFFFF
+
+        if i >= 0x80000000:
+            I = hmac.digest(self.chaincode, b'\x00' + self.sk.private_key + i2leosp(32, i), 'sha512')
+        else:
+            I = hmac.digest(self.chaincode, self.sk.pubkey.serialize(compressed=True) + i2leosp(32, i), 'sha512')
```
nuttycom, line 42: I was just checking over this again, and I realized I'm not sure whether this conforms to:
> If not (normal child): let I = HMAC-SHA512(Key = cpar, Data = serP(point(kpar)) || ser32(i)).

Does pubkey.serialize do:
> point(p): returns the coordinate pair resulting from EC point multiplication (repeated application of the EC group operation) of the secp256k1 base point with the integer p.

nuttycom: https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki contains some test vectors, perhaps we should import those and use them to verify our bip32 implementation.

daira: Yes, that's what it does. And that's a good idea (but doesn't affect this PR).

nuttycom (approved):
ACK. With the fix to index endianness, our bip32 generation conforms to the Bitcoin test vectors, and I have verified that the test vectors are correctly matched by zcashd unified address derivation in https://github.com/zcash/zcash/pull/5541

nuttycom (approved):
ACK!

PR #85, opened 2022-03-11
Add structural test vectors for transaction formats

### LINKED ISSUE #84: Add transaction parser test vectors
We have a `Transaction` type that generates random structurally-valid (but data-invalid) transactions. We use this in the sighash test vectors (e.g. for ZIP 244). It would be separately useful to have test vectors for the transaction parser itself, that breaks out random transactions into their individual pieces in a way that downstream developers can test against.

str4d:
Closes #84.


regenerate.sh
```diff
@@ -19,6 +19,7 @@ tv_scripts=(
     sapling_note_encryption
     sapling_signatures
     sapling_zip32
+    transaction_v5
```
daira, line 22: Why is `transaction_legacy` not included?

zcash_test_vectors/transaction.py
```diff
@@ -471,7 +479,11 @@ def __init__(self, rand, consensus_branch_id):
             if is_coinbase:
                 # set enableSpendsOrchard = 0
                 self.flagsOrchard &= 2
-            self.valueBalanceOrchard = rand.u64() % (MAX_MONEY + 1)
+            valueBalanceRand = rand.u64()
+            if valueBalanceRand & 0x8000000000000000:
+                self.valueBalanceOrchard = (valueBalanceRand % (MAX_MONEY + 1)) - MAX_MONEY
+            else:
+                self.valueBalanceOrchard = valueBalanceRand % (MAX_MONEY + 1)
```
daira, line 486: Three times and you factor out:
```
def signed_amount(rand):
    valueRand = rand.u64()
    if valueRand & 0x8000000000000000:
        return (valueRand % (MAX_MONEY + 1)) - MAX_MONEY
    else:
        return valueRand % (MAX_MONEY + 1)
```

zcash_test_vectors/transaction.py
```diff
@@ -563,3 +575,151 @@ def __getattr__(self, item):
...
+        ret = []
+        while len(ret) < l:
+            ret.append(rng.randrange(0, 256))
+        return bytes(ret)
+    rand = Rand(randbytes)
+
+    test_vectors = []
+    while len(test_vectors) < 30:
+        # Generate transactions with versions prior to ZIP 225.
+        tx = LegacyTransaction(rand, rand.u8() % NU5_TX_VERSION)
```
daira, line 601: 0 was never a valid version number. I see that line 327 above will replace a version other than 3 (Overwinter) and 4 (Sapling) with a random version:
```
            self.nVersion = rand.u32() & ((1 << 31) - 1)
```
but I think that versions 1 and 2 (as version numbers that were actually used) should have high probability of occurrence, and 0 should be excluded.

zcash_test_vectors/transaction.py
```diff
@@ -471,7 +479,11 @@ def __init__(self, rand, consensus_branch_id):
...
                 self.flagsOrchard &= 2
-            self.valueBalanceOrchard = rand.u64() % (MAX_MONEY + 1)
+            valueBalanceRand = rand.u64()
+            if valueBalanceRand & 0x8000000000000000:
+                self.valueBalanceOrchard = (valueBalanceRand % (MAX_MONEY + 1)) - MAX_MONEY
+            else:
+                self.valueBalanceOrchard = valueBalanceRand % (MAX_MONEY + 1)
             self.anchorOrchard = PallasBase(leos2ip(rand.b(32)))
             self.proofsOrchard = rand.b(rand.u8() + 32) # Proof will always contain at least one element
             self.bindingSigOrchard = RedPallasSignature(rand)
```
daira, line 489: Below here: `version_bytes` is confusingly named because it doesn't return `bytes`.

daira (approved):
utACK with comments.

LarryRuane (commented):
concept ACK

LarryRuane:
I'm trying to use the transaction test vectors here to test lightwalletd parsing. Maybe I'm interpreting things wrongly, but it seems like the `nConsensusBranchId` values are wrong in the broken-out fields. Immediately after the occurrences of `5, 648488714,`, which are `version` and the `nVersionGroupId` and are both correct, there are many different values: 1938782813, 626621063, 1719046631, ... Shouldn't those all be 3268858036 (=[0xc2d6d0b4](https://github.com/zcash/zcash/blob/33c813c7a17407f252adf63ff3de45af827002c1/src/consensus/upgrades.cpp#L48))?

str4d:
@LarryRuane These are structural test vectors, and structurally the `nConsensusBranchId` field may contain any value. If we forced it to always be set to the NU5 consensus branch ID, then we'd never be able to use v5 transactions in any subsequent NU.

Incidentally, this is something I plan to change about the Rust transaction parser, which currently applies the additional restriction that `nConsensusBranchId` must be some known value. That restriction is a consensus rule, but it's not actually something the parser should concern itself with (and it made running a private NU5 testnet harder).

PR #89, opened 2022-09-15
Provide UA test vectors for the first 3 valid diversifier indices for each account

nuttycom:


pacu (approved):
utACK

String-encoded UAs would be useful for testing the SDKs. Probably that change is better on another PR

pacu:
CI is failing :/

.github/workflows/test_vectors.yml
```diff
@@ -24,6 +24,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: Install gnome-keyring
+        run: sudo apt-get install gnome-keyring
```
str4d, line 28: What is this needed for?

str4d (approved):
utACK 73ced0f5c986d7dceeb231ce5bf1822bb6dc2673

.github/workflows/test_vectors.yml
nuttycom: One of the python dependencies refuses to build without it; that was the previous build failure.

PR #90, opened 2022-09-28
Ensure that successive diversified UAs have the same structure.

nuttycom:
In 23321662d1b36003ea178d3a55db7890ffb7a6d0, the unified address test vectors were modified to include multiple addresses for each account at successive diversifier indices. However, the generation process resulted in such diversified addresses having varying sets of typecodes within a single account; in particular, p2sh and p2pkh addresses were interleaved in a manner that made testing transparent functionality using the resulting test vectors more difficult.

In the process of this fix, it was also noted that the handling of unknown typecodes was inconsistent, with the unknown typecode always being included in the resulting test vector even if no data corresponding to that typecode was present. This has also been modified such that if an unknown typecode is present, data is present, and vice versa.


zcash_test_vectors/unified_address.py
```diff
@@ -39,7 +39,11 @@ def main():
 
     test_vectors = []
     for account in range(0, 20):
+        # Each set of sequential diversified addresses should have the same set
+        # of typecodes, to simplify use in tests.
         has_t_addr = rand.bool()
+        # use p2pkh 3/4 of the time
+        is_p2pkh = rand.bool() or rand.bool()
```
daira, line 46: `or` is shortcutting, which might make it more difficult to refactor this code in a way that preserves existing test vector entries that should be unaffected.
```suggestion
        is_p2pkh = any([rand.bool(), rand.bool()])
```

zcash_test_vectors/unified_address.py
```diff
@@ -48,53 +52,58 @@ def main():
         else:
             t_addr = None
 
+        has_s_addr = rand.bool()
+        has_o_addr = (not has_s_addr) or rand.bool()
+        # include an unknown item 1/4 of the time
+        has_unknown_item = rand.bool() and rand.bool()
```
daira, line 58:
```suggestion
        has_unknown_item = all([rand.bool(), rand.bool()])
```
Same rationale as above.

zcash_test_vectors/unified_address.py
```diff
@@ -108,15 +117,17 @@ def main():
             assert decoded.get('orchard') == orchard_raw_addr
             assert decoded.get('sapling') == sapling_raw_addr
             assert decoded.get('transparent') == t_addr
-            assert decoded.get('unknown') == ((unknown_tc, unknown_bytes) if unknown_bytes else None)
-            assert decoded.get('transparent') == t_addr
+            if has_unknown_item:
+                assert decoded.get('unknown') == ((unknown_tc, unknown_bytes))
```
daira, line 121:
```suggestion
                assert decoded.get('unknown') == (unknown_tc, unknown_bytes)
```

daira (approved):
utACK. I have *not* regenerated the test vectors on my machine and compared them against the ones checked in here.

nuttycom:
> utACK. I have _not_ regenerated the test vectors on my machine and compared them against the ones checked in here. Someone besides @nuttycom should do that.

CI does that these days. :)

nuttycom:
@daira the first commit of this PR now fixes a bug in the derivation of the transparent receiver for the UAs.

daira (approved):
utACK

PR #91, opened 2022-09-29
Various cleanups

daira:
* Move `orchard_vesta` to `zcash_test_vectors.orchard.vesta`.
* Fix errors due to the removal of RIPEMD-160 from default-available OpenSSL hashes.
* Add a unit test runner.
* Improve the usability of `regenerate.sh` and document it.


regenerate.sh
```diff
@@ -27,8 +29,16 @@ tv_scripts=(
     zip_0244
     zip_0316)
 
+formats="${1:-rust json zcash}"
+
 for generator in "${tv_scripts[@]}"
```
nuttycom, line 34: While we're fixing this, something that I frequently find myself doing is just regenerating a single set of test vectors, instead of regenerating them all. Now that we no longer use `$2`, maybe if that's present we could use that argument as the sole value in `tv_scripts` instead of regenerating everything?

nuttycom: The approach of using `poetry run SCRIPT_NAME` directly doesn't put the files in the expected place by itself.

nuttycom (changes requested):
Now that this has a test runner, that should be added to the CI configuration. Otherwise LGTM; I have not inspected the ripemd160 implementation.

regenerate.sh
```diff
@@ -27,8 +29,16 @@ tv_scripts=(
     zip_0244
     zip_0316)
 
+formats="${*:-rust json zcash}"
```
daira, line 32: I had a bug here which I've force-pushed to correct: `${1:-rust json zcash}` -> `${*:-rust json zcash}`.

regenerate.sh
daira: The (corrected) script takes a list of formats, so there would be a conflict unless it implemented more complicated argument parsing. You can always write `poetry run SCRIPT_NAME -t FORMAT >OUTPUT_FILE`.

daira:
This broke the test vector regeneration test. Will fix.

regenerate.sh
```diff
@@ -27,8 +29,16 @@ tv_scripts=(
     zip_0244
     zip_0316)
```
nuttycom, line 30:
```suggestion
if [ -z "$2" ];
then
    tv_scripts=(
        bip_0032
        f4jumble
        f4jumble_long
        orchard_empty_roots
        orchard_generators
        orchard_group_hash
        orchard_key_components
        orchard_map_to_curve
        orchard_merkle_tree
        orchard_note_encryption
        orchard_poseidon
        orchard_poseidon_hash
        orchard_sinsemilla
        sapling_generators
        sapling_key_components
        sapling_note_encryption
        sapling_signatures
        sapling_zip32
        unified_address
        unified_full_viewing_keys
        unified_incoming_viewing_keys
        zip_0143
        zip_0243
        zip_0244
        zip_0316)
else
    tv_scripts=($2)
fi
```

regenerate.sh
nuttycom: Ah, I see. Hm. I guess I'll just have to keep commenting out bits of this script when I need to regenerate stuff.

PR #94, opened 2023-11-22
Generate hardened-only Sapling ZIP 32 test vectors

str4d:


test-vectors/json/sapling_zip32_hard.json
```diff
@@ -0,0 +1,8 @@
+[
+    ["From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/sapling_zip32.py"],
```
nuttycom, line 2: This URI is now incorrect.

str4d: That's been incorrect for a while, both due to internal refactoring to `poetry` and the recent repository move. This should be fixed separately across the entire test vector corpus.

nuttycom (approved):
utACK. I have not attempted to regenerate these locally, but CI verifies this so it should be fine.

PR #95, opened 2023-12-01
Add key prefixing to RedJubjub

str4d:
The original draft spec for RedJubjub did not include key prefixing, and both this reference implementation and the Rust implementation were implemented following that spec. Then when key prefixing was added, the Rust code was updated at its call sites instead of inside the RedJubjub implementation.

Now that we are migrating to the `redjubjub` crate which correctly builds in key prefixing, the old test vectors can no longer be used.


daira (approved):
utACK

PR #101, opened 2024-08-22
Add ZSF transaction test vectors

tomekpiotrowski:
This adds test data generation for ZSF transactions. Simply adds `zsf_deposit` to tx id and sighash calculations.

For now ZSF deposit is developed under the ZFuture Network Update and ZFuture transaction version as the final NU and TX versions under which it'll be deployed are not decided yet.

The `librustzcash` PR that uses the generated data is yet to be pushed.


giddie:
I've updated this PR following the naming change (ZSF -> NSM).

aphelionz:
Thanks @mariopil

zcash_test_vectors/transaction.py
```diff
@@ -462,6 +477,11 @@ def __init__(self, rand, consensus_branch_id):
             # v^balanceSapling is defined to be 0.
             self.valueBalanceSapling = 0
 
+    def init_orchard(self, rand, flip_coins):
+        have_transparent_in = (flip_coins >> 0) % 2
+        have_orchard = (flip_coins >> 3) % 2
+        is_coinbase = (not have_transparent_in) and (flip_coins >> 4) % 2
```
nuttycom, line 483: Same bug, `have_transparent_in` and `is_coinbase` must be consistent for the entire transaction.

nuttycom (changes requested):
Review not yet complete, but marking with changes requested to prevent inadvertent merge.

zcash_test_vectors/transaction.py
```diff
@@ -443,6 +453,11 @@ def __init__(self, rand, consensus_branch_id):
             for _ in range((rand.u8() % 3) + 1):
                 self.vout.append(TxOut(rand))
 
+    def init_sapling(self, rand, flip_coins):
+        have_transparent_in = (flip_coins >> 0) % 2
+        have_sapling = (flip_coins >> 2) % 2
+        is_coinbase = (not have_transparent_in) and (flip_coins >> 4) % 2
```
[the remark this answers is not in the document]
mariopil: Maybe `flip_coins_result` ?

zcash_test_vectors/transaction.py
```diff
@@ -462,6 +478,9 @@ def __init__(self, rand, consensus_branch_id):
             # v^balanceSapling is defined to be 0.
             self.valueBalanceSapling = 0
 
+    def init_orchard(self, rand, flip_coins_result, is_coinbase):
+        have_orchard = (flip_coins_result >> 3) % 2
+
         # Orchard Transaction Fields
         self.vActionsOrchard = []
         if have_orchard:
```
daira, line 486: It's invalid to have 0 actions but a non-zero valueBalanceOrchard, for example, but the next line allows that. Change it as follows:
```python
for _ in range(max(1, rand.u8() % 5):
```
There is a 1 in 10 chance, for each test vector, that the old code would have created a tx that is invalid in this way. So the test vectors after that point might be disrupted (because the extra call to `OrchardActionDescription` will mutate `rand`). Technically we could avoid this but don't worry about it if it happens.

zcash_test_vectors/transaction.py
```diff
@@ -443,6 +456,9 @@ def __init__(self, rand, consensus_branch_id):
             for _ in range((rand.u8() % 3) + 1):
                 self.vout.append(TxOut(rand))
 
+    def init_sapling(self, rand, flip_coins_result, is_coinbase):
+        have_sapling = (flip_coins_result >> 2) % 2
+
         # Sapling Transaction Fields
         self.vSpendsSapling = []
         self.vOutputsSapling = []
```
daira, line 464: valueBalanceSapling is "The net value of Sapling spends minus outputs."
So it's invalid to have a positive valueBalanceSapling if there are no spends, or a negative valueBalanceSapling if there are no outputs (this is not enforced as a separate consensus rule but it holds under the assumption of soundness of the spend and output circuits). Also there cannot be a Sapling anchor when there are no Sapling spends, and there cannot be a Sapling binding signature if there are neither Sapling spends nor outputs.

Change it like this:
````python
        if have_sapling:
            # We use the randomness unconditionally here to avoid unnecessary test vector changes.
            anchorSapling = Fq(leos2ip(rand.b(32)))
            for _ in range(rand.u8() % 3):
                spend = SpendDescription(rand, self.anchorSapling)
                if not is_coinbase:
                    self.vSpendsSapling.append(spend)
            for _ in range(rand.u8() % 3):
                self.vOutputsSapling.append(OutputDescription(rand))
            valueBalanceSapling = rand.u64() % (MAX_MONEY + 1)
            bindingSigSapling = RedJubjubSignature(rand)
            if len(self.vSpendsSapling) > 0:
                self.anchorSapling = anchorSapling
            if len(self.vSpendsSapling) > 0 or len(self.vOutputsSapling) > 0:
                self.bindingSigSapling = bindingSigSapling
            if len(self.vSpendsSapling) == 0:
                valueBalance = min(0, valueBalance)
            if len(self.vOutputsSapling) == 0:
                valueBalance = max(0, valueBalance)
            self.valueBalance = valueBalance
```

zcash_test_vectors/transaction.py
```diff
@@ -549,12 +588,34 @@ def __bytes__(self):
...
         return ret
 
+class TransactionV6(TransactionV5):
+    def init_header(self, consensus_branch_id, rand):
+        # Common Transaction Fields
+        self.nVersionGroupId = V6_VERSION_GROUP_ID
+        self.nConsensusBranchId = consensus_branch_id
+        self.nLockTime = rand.u32()
+        self.nExpiryHeight = rand.u32() % TX_EXPIRY_HEIGHT_THRESHOLD
+        self.zip233Amount = rand.u64() % (MAX_MONEY + 1)
```
daira, line 598: This results in there being a negligible chance of zero burn, which is unrealistic.
```suggestion
        self.zip233Amount = rand.u64() % (MAX_MONEY + 1)
        if rand.u8() % 2 == 0:
            self.zip233Amount = 0
```

zcash_test_vectors/zip_0233.py
```diff
@@ -0,0 +1,120 @@
...
+            t_inputs = [TransparentInput(nIn, rand) for nIn in range(len(tx.vin))]
+            sum_amount = sum(x.amount for x in t_inputs)
+
+            # Be sure total amount doesn't exceed MAX_AMOUNT
+            if sum_amount > MAX_MONEY:
+                t_inputs[-1] = MAX_MONEY - sum_amount
+                sum_amount = MAX_MONEY
+
+            if tx.zip233Amount + sum_amount > MAX_MONEY:
+                tx.zip233Amount = MAX_MONEY - sum_amount
```
daira, line 50: This isn't a correct way to ensure that the sum of input amounts + the zip233Amount is $\leq$ MAX_MONEY. The amount of each input is independently determined as:
```python
        self.amount = rand.u64() % (MAX_MONEY + 1)
```
So the sum of inputs before the last might already be $>$ MAX_MONEY, in which case `t_inputs[-1]` would become negative on line 46, which is invalid.

Note that `zip_0244.py` doesn't constrain the input total to be $\leq$ MAX_MONEY, but that's ok, don't worry about it. We shouldn't duplicate that bug for new test vectors.

An adequate way to solve this is to just use a smaller range for each input. Change `TransparentInput` like this:
```python
class TransparentInput(object):
    def __init__(self, nIn, rand, limit=MAX_MONEY):
        self.nIn = nIn
        self.scriptPubKey = Script(rand)
        self.amount = rand.u64() % (limit + 1)
```
and then do something like:
```suggestion
        t_inputs = []
        if not tx.is_coinbase():
            sum_amount = 0
            nIn = len(tx.vin)
            if nIn > 0:
                t_inputs = [TransparentInput(i, rand, MAX_MONEY // (nIn-1)) for i in range(nIn-1))]
                sum_amount = sum(x.amount for x in t_inputs)
                # Ensure that at least one of the inputs can reach the full range.
                t_inputs.append(TransparentInput(nIn-1, rand, MAX_MONEY - sum_amount))
                sum_amount += t_inputs[-1].amount
       
        self.zip233Amount = rand.u64() % (MAX_MONEY - sum_amount + 1)
        # Make half the burn amounts 0 for a more realistic distribution.
        if rand.u8() % 2 == 0:
            self.zip233Amount = 0
```

daira, line 50: This isn't a correct way to ensure that the sum of input amounts + the zip233Amount is $\leq$ MAX_MONEY. The amount of each input is independently determined as:
```python
        self.amount = rand.u64() % (MAX_MONEY + 1)
```
So the sum of inputs before the last might already be $>$ MAX_MONEY, in which case `t_inputs[-1]` would become negative on line 46, which is invalid.

Note that `zip_0244.py` doesn't constrain the input total to be $\leq$ MAX_MONEY, but that's ok, don't worry about it. We shouldn't duplicate that bug for new test vectors.

An adequate way to solve this is to just use a smaller range for each input. Change `TransparentInput` like this:
```python
class TransparentInput(object):
    def __init__(self, nIn, rand, limit=MAX_MONEY):
        self.nIn = nIn
        self.scriptPubKey = Script(rand)
        self.amount = rand.u64() % (limit + 1)
```
and then do something like:
```suggestion
        t_inputs = []
        if not tx.is_coinbase():
            sum_amount = 0
            nIn = len(tx.vin)
            if nIn > 0:
                t_inputs = [TransparentInput(i, rand, MAX_MONEY // (nIn-1)) for i in range(nIn-1))]
                sum_amount = sum(x.amount for x in t_inputs)
                # Ensure that at least one of the inputs can reach the full range.
                t_inputs.append(TransparentInput(nIn-1, rand, MAX_MONEY - sum_amount))
                sum_amount += t_inputs[-1].amount
       
        self.zip233Amount = rand.u64() % (MAX_MONEY - sum_amount + 1)
        # Make half the burn amounts 0 for a more realistic distribution.
        if rand.u8() % 2 == 0:
            self.zip233Amount = 0
```

zcash_test_vectors/zip_0244.py
daira, line 1: Add a comment before line 364:
```python
            # We don't attempt to avoid generating amounts than sum to more than MAX_MONEY.
```

daira (changes requested):
Looks good modulo comments.

mariopil:
Thank you @daira for the suggestions, I've updated the PR.

zcash_test_vectors/transaction.py
```diff
@@ -26,6 +26,10 @@
 NU5_VERSION_GROUP_ID = 0x26A7270A
 NU5_TX_VERSION = 5
 
+V6_TX_VERSION = 6
+# TODO: change this
+V6_VERSION_GROUP_ID = 0xFFFFFFFF
```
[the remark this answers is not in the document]
mariopil: I've added this to match the versions in librustzcash, ensuring the generated vectors used in the librustzcash tests are correct.

daira (dismissed):
utACK, except that the checked-in test vectors need to be updated.

aphelionz:
@mariopil @daira what are the next steps?

nuttycom:
This looks like it now needs to be rebased on `main` following #109

zcash_test_vectors/zip_0233.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
+import sys; assert sys.version_info[0] >= 3, "Python 3 required."
+
+from hashlib import blake2b
+import struct
+
+from .transaction import TransactionV6
+from .output import render_args, render_tv, Some
```
oxarbitrage:
```suggestion
from .output import render_args, render_tv
```

zcash_test_vectors/zip_0233.py
```diff
@@ -0,0 +1,121 @@
...
+    args = render_args()
+
+    from random import Random
+    rng = Random(0xB7D6_0F44)
+    def randbytes(l):
+        ret = []
+        while len(ret) < l:
+            ret.append(rng.randrange(0, 256))
+        return bytes(ret)
+    rand = Rand(randbytes)
```
oxarbitrage:
```suggestion
    rand = Rand(randbytes(rng))
```

Function exist and in scope for this.

zcash_test_vectors/zip_0244.py
```diff
@@ -360,6 +361,7 @@ def randbytes(l):
         if tx.is_coinbase():
             t_inputs = []
         else:
+            # We don't attempt to avoid generating amounts than sum to more than MAX_MONEY.
```
oxarbitrage: Should we do this in this PR or what is the plan ?

oxarbitrage (commented):
Optional suggestions and some questions/comments. All minor stuff, feel free to ignore.

PR #103, opened 2024-11-02
Test vectors for ZIP 32 arbitrary key derivation

str4d:


test-vectors/json/zip_0032_arbitrary.json
```diff
@@ -0,0 +1,8 @@
+[
+    ["From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/zip_0032_arbitrary.py"],
+    ["sk, c"],
```
daira, line 3: For the other test vectors I think we normally include the input (in this case the context, and the account id in the case of account-specific key derivation), not just the output. For example see https://github.com/zcash/zcash-test-vectors/blob/master/test-vectors/rust/f4jumble.rs or https://github.com/zcash/zcash-test-vectors/blob/master/test-vectors/rust/orchard_group_hash.rs .

daira (approved):
utACK with non-blocking comment.

test-vectors/json/zip_0032_arbitrary.json
str4d: These test vectors follow how the existing ZIP 32 test vectors work for Sapling and Orchard.

PR #104, opened 2024-11-13
Improve the ZIP 32 arbitrary key derivation test vectors

str4d:
When using them, I forgot that `bytes(range(32))` doesn't generate an all-zeroes seed of length 32, so I was using the wrong seed with the test vectors and they were failing. We now embed that seed in the test vectors, along with the context string (to make it clear that there is no trailing null byte) and the IKM (which is what clued me into the seed being wrong).


zcash_test_vectors/zip_0032.py
```diff
@@ -71,10 +71,19 @@ def arbitrary_key_derivation_tvs():
     m_1h_2h = m_1h.child(hardened(2))
     m_1h_2h_3h = m_1h_2h.child(hardened(3))
 
-    keys = [m, m_1h, m_1h_2h, m_1h_2h_3h]
+    # Derive a path matching Zcash mainnet account index 0.
+    m_32h = m.child(hardened(32))
```
nuttycom, line 75: Nit: it would be nice to link to https://zips.z.cash/zip-0032#key-path-levels here. Also I note that ZIP 32 doesn't explicitly mention coin type 133 in this section, it probably should.

nuttycom (approved):
utACK with doc suggestion

PR #105, opened 2025-02-17
Add test vectors for ZIP 32 registered key derivation

daira:


zcash_test_vectors/zip_0032.py
```diff
@@ -26,23 +26,28 @@ def MKGh(Context, IKM):
...
 
-def CKDh(Context, sk_par, c_par, i):
+def CKDh(Context, sk_par, c_par, i, tag, full_width_leaf):
     assert type(Context) == HardenedOnlyContext
     assert 0x80000000 <= i and i <= 0xFFFFFFFF
+    assert type(tag) == bytes
+    assert type(full_width_leaf) == bool
 
-    I   = prf_expand(c_par, Context.CKDDomain + sk_par + i2leosp(32, i))
+    leaf = b"" if tag == b"" and not full_width_leaf else int(full_width_leaf).to_bytes()
```
str4d, line 35: This needs to be updated to match the updated spec.

zcash_test_vectors/zip_0032.py
```diff
@@ -53,12 +58,62 @@ def master(cls, ContextString, S):
...
+            ('subpath',    '&\'static [(u32, &\'static [u8])]'),
+            ('sk',         '[u8; 32]'),
+            ('c',          '[u8; 32]'),
+            ('full_width', 'Option<[u8; 64]>'),
+        ),
+        test_vectors,
+    )
+
+
+class ArbitraryKey(RegisteredKey):
```
str4d, line 115: This should be reverted; ZIP 32 does not have ArbitraryKey as a sub-instance of RegisteredKey, and the intention of the test vector implementations is to be as close to the specifications as possible (vs the production impls which may take liberties with APIs for performance reasons).

zcash_test_vectors/zip_0032.py
```diff
@@ -26,23 +26,28 @@ def MKGh(Context, IKM):
...
     I_L = I[:32]
     I_R = I[32:]
     return (I_L, I_R)
 
-class ArbitraryKey(object):
-    Arbitrary = HardenedOnlyContext(b'ZcashArbitraryKD', b'\xAB')
 
-    def __init__(self, IKM, path, sk, chaincode):
+class RegisteredKey(object):
+    Context = HardenedOnlyContext(b'ZIPRegistered_KD', b'\xAC')
```
str4d, line 43: After the reversion, this would then become (for spec closeness):
```suggestion
    Registered = HardenedOnlyContext(b'ZIPRegistered_KD', b'\xAC')
```

zcash_test_vectors/zip_0032.py
```diff
@@ -26,17 +26,94 @@ def MKGh(Context, IKM):
...
+        return cls(IKM, [], sk, chaincode)
+
+    def child(self, i):
+        (sk_child, c_child) = CKDh(self.Registered, self.sk, self.chaincode, i, 0, b"")
+        return self.__class__(None, self.path + [i], sk_child, c_child)
+
+    def child_with_tag(self, i, tag):
+        (sk_child, c_child) = CKDh(self.Registered, self.sk, self.chaincode, i, 0, tag)
+        (left, right) = CKDh(self.Registered, self.sk, self.chaincode, i, 1, tag)
+        return self.__class__(None, self.path + [(i, tag)], sk_child, c_child, left + right)
```
str4d, line 71: - The ZIP does not expose the ability to generate the tree root directly, so having it in test vectors would encourage potentially unwanted production APIs.
- There's no explicit `zip_number` field in the API matching the spec; instead the caller is expected to pre-harden the ZIP number and insert it in the correct place in the path.
- The spec does not have separate `child` and `child_with_tag` APIs; that's an artifact of the Rust implementation.

Instead, these methods should match the spec:
```suggestion
    @classmethod
    def master(cls, ContextString, S, ZipNumber):
        length_ContextString = len(ContextString)
        length_S = len(S)

        assert length_ContextString <= 252
        assert 32 <= length_S <= 252

        IKM = bytes([length_ContextString]) + ContextString + bytes([length_S]) + S
        (sk_m, c_m) = MKGh(cls.Registered, IKM)
        i = hardened(ZipNumber)
        (sk, chaincode) = CKDh(self.Registered, sk_m, c_m, i, 0, b"")
        return cls(IKM, [i], sk, chaincode)

    def child(self, i, tag):
        (sk_child, c_child) = CKDh(self.Registered, self.sk, self.chaincode, i, 0, tag)
        (I_L, I_R) = CKDh(self.Registered, self.sk, self.chaincode, i, 1, tag)
        return self.__class__(None, self.path + [(i, tag)], sk_child, c_child, I_L + I_R)
```

daira:
Ready for re-review (commit-by-commit suggested).

str4d (approved):
utACK cbfe5ed8280cbcc7ff8f39dc3ae3680287e4bdd8. Thanks for refactoring the commits ❤️

PR #107, opened 2025-07-16
Add ZIP 32 Seed Fingerprint encoding to test vectors

str4d:


nuttycom (approved):
ACK. I have not verified the generated values.

PR #130, opened 2026-06-18
Add Ironwood v6 transaction hash vectors

czarcas7ic:
Adds an Ironwood v6 transaction hash vector generator and generated fixtures for txid, auth digest, per-pool digests, and shielded plus transparent sighashes.

The shared transaction hash helper extraction keeps the existing ZIP 244 and ZIP 233 generators on the same code path, while the new generator covers v6 Sapling, Orchard, and Ironwood anchor movement plus Ironwood flag behavior. The branch id remains the provisional 0xffffffff value used by the current NU6.3 implementation.

Validation:
- uv run --python python3.11 ./regenerate.sh all ironwood_v6_tx_hash
- uv run --python python3.11 ./regenerate.sh all zip_0244
- uv run --python python3.11 ./regenerate.sh all zip_0233
- git diff --check master...HEAD
- imported the generated Rust fixture into the current librustzcash Ironwood implementation and ran the v6 tx hash vector test under zcash_unstable="nu6.3"


test-vectors/rust/ironwood_v6_tx_hash.rs
```diff
@@ -0,0 +1,6928 @@
+// From https://github.com/zcash/zcash-test-vectors/blob/master/zcash_test_vectors/ironwood_v6_tx_hash.py
```
czarcas7ic, line 1: This is generated from `uv run ironwood_v6_tx_hash -t rust | rustfmt --edition 2021`

zcash_test_vectors/ironwood_v6_tx_hash.py
```diff
@@ -0,0 +1,281 @@
...
+from .sapling.jubjub import Fq
+from .zip_0143 import (
+    SIGHASH_ALL,
+    SIGHASH_ANYONECANPAY,
+    SIGHASH_NONE,
+    SIGHASH_SINGLE,
+)
+
+
+IRONWOOD_EXPERIMENTAL_BRANCH_ID = 0xFFFFFFFF
```
czarcas7ic, line 33: Needs to change once branch ID is selected.

czarcas7ic:
Imported the rust test file from this PR into our librustzcash PR and ran

`RUSTFLAGS='--cfg zcash_unstable="nu6.3"' cargo test -p zcash_primitives ironwood_v6_tx_hash_vectors --no-default-features --features test-dependencies`

And it passed, giving me more confidence we have parity between the two.

